PR #171, opened 2022-03-13
fix: design feedback

nicoglennon:
a bunch of small UI fixes for the feedback from jeremy & mikey


peduarte (approved):
left some small comments, but otherwise it's looking great!

PR #177, opened 2022-03-16
chore: new colors

mikearndt:
added the color spectrums and finished adding the semantic tokens.


peduarte (approved):
Thanks for populating these dude!

PR #191, opened 2022-03-23
feat: make Chain metadata more versatile, switch our explorers for useNetwork's instead

nicoglennon:
i started this PR by turning the `ChainIconsContext` into a more general `RainbowKitChainsContext` in order to be able to add the `explorerUrl` prop alongside the `iconUrl`. once i had it working i accidentally realized that wagmi's `Chain` actually provides an array of explorers for each chain, so we can use the already provided useNetwork instead of throwing it in our context (developers can provide this themselves during chain setup if configuring custom chains / not default ones). i decided to leave the refactor of ChainIconsContext to RainbowKitChainsContext in case that in the future we want to add another property internally, we can do so more easily.

using kovan custom chain: 
https://user-images.githubusercontent.com/16931094/159797925-fd76a21e-7aa1-48c2-b573-6eff6efc106b.mov


markdalgleish (approved):
Just added one small comment about types.

This is a really nice clean up 🙌

PR #194, opened 2022-03-24
feat: add new theme colors for light dark

nicoglennon:
added the new accent colors for light & dark mode 

Screen Shot 2022-03-24 at 1 33 09 AM


markdalgleish (approved):
This should probably have a changeset to at least cover the new accent colors, but this looks great 👍

PR #199, opened 2022-03-28
feat: add prototype of transitions API

markdalgleish:
There's a bit here to discuss so I've opted for a draft PR for now to get the ball rolling.

As an example of how this might work, I've added a new `transitions` option on the built-in themes that optionally accepts either `"large"` or `"small"`. Of course all of these names and values are just placeholders for now to give us something to help frame API discussions and play around with.

Note that there are two layers to this API that need discussion:

First, the low-level theme API that gives fine grained control over all notable transitions, which currently looks like this:

```
transitions: {
  modalEntrance: {
    fadeDuration: '...',
    slideDistance: '...',
    slideDuration: '...',
    slideEasing: '...',
  },
  modalEntranceMobile: {
    fadeDuration: '...',
    slideDistance: '...',
    slideDuration: '...',
    slideEasing: '...',
  },
},
```

Secondly, the high-level API for built-in themes, similar to how we handle `borderRadius`, for example:

```
const theme = darkTheme({
  accentColor: 'purple',
  transitions: 'small',
});
```


markdalgleish:
Closing for now. We may revisit this in the future.

PR #203, opened 2022-03-29
fix: drop switch button for now

nicoglennon:
Screen Shot 2022-03-29 at 1 52 33 AM


nickbytes (approved):
nit on removal vs comment, but otherwise it's all good.

PR #205, opened 2022-03-30
fix: prevent body from scrolling while modal is open

markdalgleish:


nickbytes (approved):
Never used react-remove-scroll, looks nice!

PR #207, opened 2022-03-30
fix: update various styles

mikearndt:
- Smaller info icons for table
- Heavier code font weight
- Shadow and transition effect to left rail menu
- Layouts changes on: li, h2, h3, etc
- New li style


peduarte (approved):
siiick

PR #210, opened 2022-03-31
feat: inline wallet icons as data URLs

markdalgleish:
Running these through [esbuild's dataurl loader](https://esbuild.github.io/content-types/#data-url) so we can manage the image files on disk.

Note that I've added a `global.d.ts` file to `packages/rainbowkit/src` so TypeScript can understand image imports.


nicoglennon (approved):
wow really nice solution 👏👏👏

PR #215, opened 2022-03-31
fix: add hardhat iconurl

nicoglennon:
Screen Shot 2022-03-31 at 3 07 59 AM


markdalgleish (approved):
This is probably worth a changeset—what do you think? Otherwise looks good 🙌

PR #217, opened 2022-04-04
feat: defer loading wallet/chain icon data URLs until mount

markdalgleish:
This PR also bundles chain icons as Base64 data URLs.

Note that this only downloads the icons that are required based on the configured chains and wallets. Here's a screenshot of what it looks like in dev tools after `preloadImages` is called inside `ConnectButtonRenderer`:

Icon chunks

In the future we may want to provide more fine-grained control over image preloading logic, but doing it automatically on mount is a good starting point and solves the main bundle size problem when including Base64 URLs in the package.


nicoglennon (approved):
u snapped

PR #233, opened 2022-04-06
fix: remove inner background if no balance shown

nicoglennon:
Screen Shot 2022-04-06 at 4 44 16 PM


markdalgleish (approved):
Just a small naming nit, but otherwise 🚀

PR #237, opened 2022-04-07
fix: weird background changes on mobile

nicoglennon:
fix this bizzare mobile only css bug. shoutout @pugson 🤴

Image from iOS


pugson (approved):
lfg 🔥

markdalgleish (approved):

PR #238, opened 2022-04-07
fix: make asyncimage work with hardocded colors

nicoglennon:
fix qr border after asyncimage updates

Screen Shot 2022-04-07 at 6 05 07 PM


markdalgleish (approved):
One comment about naming, but otherwise looks good 👍

PR #243, opened 2022-04-11
fix: explicit textAlign center

nicoglennon:
fixes a mobile bug i introduced on friday trying to normalize textAlign

@markdalgleish thoughts on the `inherit` default for the `Text` component? Since `Text` wraps `Box` this was one way I could get `Text` components to actually inherit from a parent with `textAlign="center"`


markdalgleish (approved):
One question about types, but otherwise looks good 👍

PR #249, opened 2022-04-14
feat: hide network switcher when using a single chain

markdalgleish:
This PR makes the following improvements:
- The network switcher is only visible when multiple chains have been configured.
- The "Invalid network" button is now labelled "Wrong network" (same as how Uniswap labels it) since it's not really an issue of validity.
- Only the "Wrong network" button is visible when on an unsupported chain. This is particularly important on mobile to reduce the screen real estate used, and also decouples it from the network switcher since it's optional.

Here's a video of the user flow when connecting to an app with a single chain configured using a wallet that's on an unsupported chain.

https://user-images.githubusercontent.com/696693/163316373-f620682f-fb19-4f5a-84a5-97f4c98b4234.mov


nickbytes (approved):
Nice!

PR #250, opened 2022-04-14
feat: truncate ENS names

nickbytes:
RNBW-3068

- Truncates ENS names


markdalgleish (approved):
lgtm 🚀

There's obviously lots of edge cases we could look into, but this is a great start.

PR #251, opened 2022-04-15
feat: adds example actions

nickbytes:
This PR adds some common actions to ensure functionality is working between RainbowKit and wagmi. In the future, this will help in debugging users' issues, as it will provide minimal reproducible cases where basic functionality is/isn't working.

- Adds Send Transaction
- Adds Sign Message
- Adds Sign Typed Data

We should add some of the following to the `rainbowkit-examples` repo, especially for different chains.


markdalgleish (approved):
One question about inlining the `useSignTypedData` objects, but otherwise looks good 👍

PR #252, opened 2022-04-15
docs: initial implementation

peduarte:
This PR is our initial implementation of the docs. 
- support for `.mdx` files. 
- mapping of mdx elements to React components
- custom docs components such as pagination, props table, code blocks, etc.
- a mini design system built with vanilla extract
- light/dark mode
- support for atoms via sprinkles

Note: The asset on the Introduction page will be changed once we prepare a new one.

Note: I've done a self review, but it's quite late so I'll go over it again on Monday.

Note: I tried to follow our eslint rules as much as possible. I think in some situations it has made the the code slightly harder to parse. Especially around the ordering of object keys and props. In these cases, I've used a few `eslint-disable`.

Note: I've removed all files related to the homepage, because all of that will change.

---

@markdalgleish a special note for you 😗 I based the DS off of Mirror's Degen DS and some parts of the Rainbowkit package. I left some inline comments about some things. The idea is that soon we'll start working on our official DS for Web products. This is kinda a "proof of concept" for it. So I didnt spend much time making sure the DS here is super flexible, as it is only ever gonna because its only gonna be used for the docs/homepage. But still would love your feedback on the usage of vanilla extract, types, etc. Thanks!

--- 

Resolves RNBW-3307
Resolves RNBW-3308


nickbytes (approved):
:chefs-kiss:

PR #258, opened 2022-04-20
feat: load all illustrations asynchronously

markdalgleish:
This PR moves our large illustration SVGs out of JSX and into separate SVG files that are loaded asynchronously (as data URLs in separate JS chunks, the same way we handle wallet/chain icons). While converting back to SVG files I also ran them through [SVGO](https://jakearchibald.github.io/svgomg) to reduce the file size.

Doing this in isolation would normally introduce a slight delay while the image downloaded on first render, so this PR also introduces some logic to preload the images. In the case of the "create" and "scan" illustrations, the preloading is only triggered once the user is part-way through the app installation instructions, meaning that a lot of users will never even download them.

A really nice bonus is that this removes a bunch of code from mobile since the illustrations are only ever used on desktop. If you open dev tools while emulating mobile you can see the difference.


peduarte (approved):
very cool

PR #263, opened 2022-04-21
rainbowkit marketing page

peduarte:
Resolves RNBW-3310

To do still (can be done in another PR once we decide)
- Add a link to the `share feedback with us` button. for this I'm thinking of using GitHub Discussions. thoughts?
- Update assets
- Update social image


markdalgleish (approved):
🔥

PR #264, opened 2022-04-21
fix: modal border width in light mode

peduarte:
Resolves RNBW-3357

More info on the ticket above ☝️ 

Please let me know if I am overlooking this.

---

Update: based on Nico's feedback and historical context, I've:
- harded 1px border width on modal
- set the modal border color in light modal to `transparent`. this basically give us the same visual impact as having no border, but without the modal changing computed height across different themes


markdalgleish (approved):
One comment about adding a changeset, but otherwise looks good 👍

PR #266, opened 2022-04-23
feat: API Providers (& removal of infura dependency)

jxom:
wagmi 0.3 no longer relies on an `infuraId` in the examples, so i think it would be nice if rainbowkit could be api provider agnostic too (have the ability to choose between alchemy, infura, or something else).

the gist of the API can be seen [here](https://rainbowkit-site-jf4a7wj8a-rainbowdotme.vercel.app/docs/api-providers).

note: creating this PR against the `jxom/wagmi-next` branch


peduarte (approved):
very nice!

markdalgleish (approved):
Wooooooooow. This is good stuff.

PR #268, opened 2022-04-26
docs: add documentation for all built-in wallets

markdalgleish:
The full suite of wallets provided in RainbowKit weren't documented, so this PR adds some basic documentation for the API of each wallet, plus some guidance on how to use the Injected Wallet fallback effectively.

Along the way I noticed that the Injected Wallet had an unused `infuraId` property on its type signature, so I cleaned that up too.

These docs will need to be updated in/after https://github.com/rainbow-me/rainbowkit/pull/266, but this gives us an initial structure we can easily iterate on.

### DIFF
--- a/README.md
+++ b/README.md
@@ -778,12 +778,136 @@ const wallets: WalletList = [
           chain.mainnet.rpcUrls[0],
       }),
       wallet.metaMask({ chains, infuraId }),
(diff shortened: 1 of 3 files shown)


peduarte (approved):
🚀

nickbytes (approved):
Awesome

PR #269, opened 2022-04-26
feat: implement more wallets design for WC wallets

nicoglennon:
Screen Shot 2022-04-27 at 2 16 11 PM


nickbytes:
I'm going to close this for repo cleanliness/organization, but I won't delete the branch. We're going to be working on better specification for adding the long tail of wallets from the WalletConnect modal. This work was super helpful to explore how we might do this, I imagine we'll reference it again.

PR #271, opened 2022-04-27
fix: reinstate install prompt

markdalgleish:
**Note: This PR is against the `wagmi-next` branch.**

I noticed that clicking on MetaMask wasn't showing the install prompt anymore if I didn't have the extension enabled. Turns out it's because we now only switch tabs in the `onConnecting` callback which is never fired when `wallet.ready` is false.


jxom (approved):
Looks good!

PR #272, opened 2022-04-27
fix: if chain switching is not support by mobile wallet, show message to end user

nickbytes:
- Some wallets do not support programmatic chain switching
  - to check for this, we need to see if `switchNetwork` (from wagmi) is `undefined`
  - if so, we need to show a message to users in the chain switch modal that their wallet is not able to switch chains
  - this feels like a better approach than hiding the chain switching modal, because users might want to know their wallet is not using all possible capabilities.
- need better designed state here


nicoglennon (approved):
my lord and savior

PR #274, opened 2022-04-28
feat: return connectors from getDefaultWallets

markdalgleish:
In order to streamline the setup process, the `getDefaultWallets` function now returns an object containing both `connectors` and `wallets` properties. This means that most consumers will no longer need to use the `connectorsForWallets` function, accessing the generated `connectors` value instead.

This is what the cleanup looks like:

```diff
-import { getDefaultWallets, connectorsForWallets } from '@rainbow-me/rainbowkit';
+import { getDefaultWallets } from '@rainbow-me/rainbowkit';

-const wallets = getDefaultWallets({
+const { connectors } = getDefaultWallets({
  /* ... */
});

-const connectors = connectorsForWallets(wallets);
```

The wallet list is still available if you were modifying it in any way:

```diff
-const wallets = getDefaultWallets({
+const { wallets } = getDefaultWallets({
  /* ... */
});
```

### DIFF
--- a/.changeset/chilled-jobs-buy.md
+++ b/.changeset/chilled-jobs-buy.md
@@ -1,0 +1,34 @@
+---
+'@rainbow-me/rainbowkit': patch
+---
+
+Make `getDefaultWallets` return an object containing `connectors` and `wallets`
+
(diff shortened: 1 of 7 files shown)


nicoglennon (approved):
🧼

PR #283, opened 2022-05-03
chore: add jsx to lint and format script extension list

markdalgleish:
We use a JSX file for the MeshGradient component on the docs site and I noticed that lint errors weren't getting picked up until CI because JSX wasn't in the file extension list for our lint/format scripts.


nickbytes (approved):
good catch!

PR #286, opened 2022-05-03
chore: adds issue and discussion templates

nickbytes:
- Adds issue templates (used wagmi template)
- Adds discussion templates (again, used wagmi as a template because I like how they've separated these)


markdalgleish (approved):
Nice!

PR #290, opened 2022-05-03
docs: general updates

peduarte:
Resolves RNBW-3504 
Resolves RNBW-3465
Resolves RNBW-3491

---

Just some updates, simplification in the copy, and a new page for creating a custom chain.


markdalgleish (approved):
Nice!

PR #294, opened 2022-05-03
chore: update svgs

nicoglennon:
updates the svgs we use, request from mikey

Screen Shot 2022-05-03 at 6 37 30 PM
.


peduarte (commented):
this svg looks weird, im guessing it's not on purpose. im just looking at the PR changes, not rainbowkit itself

CleanShot 2022-05-04 at 01 04 05@2x

nickbytes (approved):

PR #303, opened 2022-05-04
fix: types on transaction store context

nickbytes:
- when running `release:test` we get an error for missing types on `provider`
- adds `BaseProvider` type from ethers
- confirmed `release:test` typechecking passes now, and example runs


peduarte (approved):
LGMT 🌈 🌈 🌈 🌈 🌈 🌈 🌈 🌈

PR #307, opened 2022-05-05
fix: use ethers peer dependency for providers

markdalgleish:
We currently have a direct dependency on `@ethersproject/providers`, but this means that provider instances generated by RainbowKit can potentially clash with the version of Ethers.js installed alongside wagmi. To avoid this, we’re now relying on the `ethers` peer dependency that wagmi already requires, ensuring that there's only ever a single copy of `ethers` between them.


peduarte (approved):
lfg

PR #330, opened 2022-05-06
Expose Avatar component

peterferguson:
Exposing the `Avatar` would allow dapps to have consistent avatar usage throughout their application
I made some comments about potential routes for this in discussion #328


peterferguson:
If you are happy with these change I can work on updating the docs too 👍

nicoglennon:
hey @peterferguson 🤠 thanks a lot for the feedback & PR, we agree that exposing the avatar makes sense for people trying to be consistent throughout their UI. right now the Avatar component could lead to UI issues if exposed though, since we currently don't guard for the disconnected state (this component is currently never rendered if you're disconnected). your PR and comments kicked off an internal convo about the avatar and how to best expose it in a simple but flexible way, and we think it will require a little more work on our side to do that. i'll close your PR for now but will keep you updated what we have planned 👍

PR #343, opened 2022-05-08
docs: make cool mode events passive

harshjv:
Since the script doesn't call `preventDefault`, switching to `passive` event listeners seem to be improving the animations.


markdalgleish (approved):

markdalgleish:
Thank you!

PR #349, opened 2022-05-09
fix: ensure click events are triggered on button edges

### LINKED ISSUE #317: [bug] Wallet buttons edge unclickable due to scale transform

### RainbowKit Version

0.1.0

### wagmi Version

0.3.2

### Current Behavior

A minor issue, but due to the wallet buttons scaling to 0.95 on click (`:active`), clicking on the very edges won't always work.

In theory, this could be solved by moving the relevant `button` styles (background, padding, hover/active) to the child `div` instead.

### Expected Behavior

Allow buttons clicks anywhere.

### Steps To Reproduce

Tested on the RainbowKit landing page.

_No response_

### Anything else?

https://user-images.githubusercontent.com/15910702/167036935-6fc52cc8-e74c-4e89-af5c-9b0d88ee5718.mov

markdalgleish:
Fixes #317.

To fix the issue, all scale transforms are now applied to child elements of buttons/links rather than the interactive elements themselves. This ensures that hit areas remain consistent during transforms.

One additional case that needed to be handled is when elements become larger on hover, increasing their hit area. If the user moves their mouse to the edge of the enlarged hit area and then attempts to click, the click event isn't fired. To fix this case, this PR also introduces a set of CSS utils called `increaseHitAreaForHoverTransform` which has a class for each transform value, e.g. `increaseHitAreaForHoverTransform.grow`. This class is applied to the containing button/link element so its hit area becomes larger when it is active. This is done using an absolutely-positioned pseudo element that sits outside the bounds of the original hit area using the same scale as the nested element.

I resisted the urge to create an abstraction for this since it's not clear to me it would be an improvement, running the risk of obfuscating what's happening in the DOM. We may revisit this in the future.

### DIFF
--- a/packages/rainbowkit/src/components/ConnectOptions/ConnectDetails.tsx
+++ b/packages/rainbowkit/src/components/ConnectOptions/ConnectDetails.tsx
@@ -1,5 +1,5 @@
 import React, { ReactNode, useEffect } from 'react';
-import { increaseHitAreaOnActive } from '../../css/increaseHitAreaOnActive.css';
+import { increaseHitAreaForHoverTransform } from '../../css/increaseHitAreaForHoverTransform.css';
 import { useWindowSize } from '../../hooks/useWindowSize';
 import { isSafari } from '../../utils/browsers';
 import { InstructionStepName } from '../../wallets/Wallet';
@@ -459,7 +459,7 @@ export function InstructionDetail({
         />
         <Box
           as="a"
-          className={increaseHitAreaOnActive.grow}
+          className={increaseHitAreaForHoverTransform.grow}
           display="block"
           href={wallet?.qrCode?.instructions?.learnMoreUrl}
           rel="noreferrer"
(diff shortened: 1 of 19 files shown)


packages/rainbowkit/src/components/MenuButton/MenuButton.css.ts
```diff
@@ -1,25 +1,6 @@
 import { style } from '@vanilla-extract/css';
-import { sprinkles } from '../../css/sprinkles.css';
 
-export const MenuButtonClassName = style([
```
markdalgleish, line 4: As part of an earlier solution attempt, I moved these styles to be Box props rather than classes so I was able to inspect the transform being applied. I decided this was worthwhile clean up so I left it as part of this PR.

packages/rainbowkit/src/components/ConnectOptions/ConnectDetails.tsx
```diff
@@ -96,18 +88,14 @@ export function GetDetail({
                     </Text>
                   </Box>
                 </Box>
-                <Box
-                  display="flex"
-                  flexDirection="column"
-                  gap="4"
-                  onClick={() => hasMobileCompanionApp && getMobileWallet(id)}
-                  {...(!hasMobileCompanionApp && downloadUrls?.browserExtension
-                    ? linkProps(downloadUrls.browserExtension)
```
markdalgleish, line 105: I noticed that in some scenarios we were rendering an `a` element around a `button`, so this PR also cleans this up, passing the link props to `ActionButton` instead.

nicoglennon:
awesome fix for this! 🤠

i would also add the border radius of the inner buttons to the containers to avoid this:

Image from iOS

nicoglennon (approved):

PR #359, opened 2022-05-10
fix: avatar not centered background color

nicoglennon:
fix avatar centering / bg issue 

before / after

Screen Shot 2022-05-10 at 3 14 52 PM . Screen Shot 2022-05-10 at 3 13 11 PM

### DIFF
--- a/packages/rainbowkit/src/components/Avatar/Avatar.tsx
+++ b/packages/rainbowkit/src/components/Avatar/Avatar.tsx
@@ -36,7 +36,7 @@ export function Avatar({ address, imageUrl, loading, size }: AvatarProps) {
         overflow="hidden"
         position="absolute"
         style={{
-          backgroundColor,
+          ...(!imageUrl && { backgroundColor }),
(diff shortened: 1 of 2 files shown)


packages/rainbowkit/src/components/Avatar/Avatar.tsx
```diff
@@ -53,13 +53,17 @@ export function Avatar({ address, imageUrl, loading, size }: AvatarProps) {
...
             width="full"
           />
-        ) : null}
-        {emoji}
+        ) : (
+          emoji
+        )}
       </Box>
-      {typeof loading === 'boolean' && (
+      {typeof loading === 'boolean' && loading && (
```
nicoglennon, line 66: @markdalgleish i assume we also want this check for `loading === true`, unless i'm missing something in this check -- the spinner was getting rendered always but had 0 opacity

markdalgleish (approved):
Apart from rolling back the `loading` change for now, this lgtm 👍

PR #360, opened 2022-05-11
fix: increase CSS specificity to avoid style clashes

markdalgleish:
This PR makes the following changes:

- Consolidates the `data-rk` and `data-rk-id` data attributes (used for providing CSS variable definitions) into a single `data-rk` attribute so we have a consistent root-level attribute to target.
- Uses [postcss-prefix-selector](https://www.npmjs.com/package/postcss-prefix-selector) to prepend `[data-rk]` to all styles, e.g. `.foo` becomes `[data-rk] .foo`.
- Refactors the esbuild config so that plugins are defined in the same file. We had a separate plugins file, but this was leftover from when we had multiple packages.


peduarte (approved):
Nice 👌

PR #368, opened 2022-05-11
chore: update wagmi to fix MetaMask in brave

nickbytes:
- Updates wagmi to get new connector usage: https://github.com/tmm/wagmi/pull/436
- Testing in Brave + Chrome w/ Metamask
- Smoke tested other wallets.

### DIFF
--- a/packages/example/package.json
+++ b/packages/example/package.json
@@ -12,7 +12,7 @@
     "react-dom": "^18.0.0"
   },
   "peerDependencies": {
-    "wagmi": "^0.3.0"
+    "wagmi": "^0.3.5"
   },
   "scripts": {
     "dev": "next dev",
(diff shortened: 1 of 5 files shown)


packages/example/package.json
```diff
@@ -12,7 +12,7 @@
     "react-dom": "^18.0.0"
   },
   "peerDependencies": {
-    "wagmi": "^0.3.0"
+    "wagmi": "^0.3.5"
```
markdalgleish, line 15: Since RainbowKit doesn't have any changes and the latest version of wagmi is still satisfied by our version range, we should probably leave this off. What do you think?

markdalgleish: Sorry, this comment is meant to be for this line: https://github.com/rainbow-me/rainbowkit/pull/368/files#diff-ab652f6c376cf926bae2d5acb6eaeae51dfc8c6dc57dfd21eb0d3e98fc1001cdR45

nickbytes: Updated

markdalgleish (approved):

PR #378, opened 2022-05-13
feat: animations for modal only

nicoglennon:
Transitions V1 as per @markdalgleish 's previous work + adding `fade` and `none`, for no modal transition.

I want to discuss with @markdalgleish a way to expand this to every interaction while still being able to use our convenient`transform` API for the `Box` component. I spend a long time yesterday trying different ways to redo what is being done here for the modal fade-in for other pieces but since these rely on the Box API and are not CSS variables I couldn't quite figure out how to hook them up properly.


packages/example/pages/_app.tsx
```diff
@@ -83,11 +83,16 @@ type AccentColor = typeof accentColors[number];
 const radiusScales = ['large', 'medium', 'small', 'none'] as const;
 type RadiusScale = typeof radiusScales[number];
 
+const transitionScales = ['large', 'small', 'fade', 'none'] as const;
```
markdalgleish, line 86: I think `fade` should be renamed to `small` because this API is likely to be used for transitions other than the modal. If so, should `small` be renamed to `medium`?

markdalgleish:
This will need a changeset and docs — or was this meant to be undocumented until the feature is expanded further?

nicoglennon:
yes I should change this to draft until the thing is done, sorry

PR #379, opened 2022-05-13
feat: add label prop to ConnectButton

nicoglennon:
Screen Shot 2022-05-13 at 1 33 25 PM

### DIFF
--- a/site/data/docs/connect-button.mdx
+++ b/site/data/docs/connect-button.mdx
@@ -21,10 +21,17 @@ export const YourApp = () => {
 
 ### Props
 
-The `ConnectButton` component exposes several props to customize its appearance by toggling the visibility of different elements.
+The `ConnectButton` component exposes several props to customize its appearance, including toggling the visibility of different elements.
 
 <PropsTable
   data={[
+    {
+      name: 'label',
+      required: false,
+      default: 'Connect Wallet',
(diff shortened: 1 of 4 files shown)


markdalgleish (approved):

site/data/docs/connect-button.mdx
```diff
@@ -59,6 +66,16 @@ The `ConnectButton` component exposes several props to customize its appearance
 
 ### Examples
 
+#### Label
+
+By default your Connect Button will say "Connect Wallet".
```
peduarte, line 71: in order to keep the doc copy as minimal as possible, I think we could remove this line since it's already documented in the props table above

site/data/docs/connect-button.mdx
```diff
@@ -59,6 +66,16 @@ The `ConnectButton` component exposes several props to customize its appearance
...
+
+For a custom label, pass your string to the `label` prop.
```
peduarte, line 73: could we use an assertive tone, for consistency with the rest of the docs?

something along the lines of: Use the `label` prop to set a custom `ConnectButton` text.

nicoglennon: for sure, sorry i know nothing about this kind of stuff

PR #389, opened 2022-05-17
feat: add `onConnectChange` props to ConnectButton

samyouxyz:
### Problem
- cannot trigger custom function when wallet connected or disconnected

### Solution
- add `onConnectChange` as props to ConnectButton

### Usage

When wallet connects.
```tsx
<ConnectButton
  onConnectChange={({ isConnected, data }) => {
    isConnected && console.log('gm!');
  }}
/>
```

When wallet disconnects.
```tsx
<ConnectButton
  onConnectChange={({ isConnected, data }) => {
    !isConnected && console.log('gn!');
  }}
/>
```

### More info
**Type of params from `onConnectChange` callback is `ConnectionInfo`**
```
type ConnectionInfo = {
  isConnected: boolean;
  data?: ConnectorData | undefined;
};

// Imported from wagmi
type ConnectorData<Provider = any> = {
    account?: string;
    chain?: {
        id: number;
        unsupported: boolean;
    };
    provider?: Provider;
};
```

The params from the callback like `data` can also be useful in cases such as logging analytics. Thanks @peduarte for this feedback.


nickbytes:
@samyouxyz thanks for the PR! This is a cool concept, I can see how it would be useful. 

Could the same thing be achieved by watching for changes in [useAccount's account](https://wagmi.sh/docs/hooks/useAccount) or [useConnect's active connector](https://wagmi.sh/docs/hooks/useConnect)?

samyouxyz:
> @samyouxyz thanks for the PR! This is a cool concept, I can see how it would be useful.
> 
> Could the same thing be achieved by watching for changes in [useAccount's account](https://wagmi.sh/docs/hooks/useAccount) or [useConnect's active connector](https://wagmi.sh/docs/hooks/useConnect)?

@nickbytes thanks for the comment!

Actually, using `useAccount` with useEffect to watch the change was my first attempt to achieve this. But I hit a use case where depending on state changes becomes unpredictable and doesn't cover all situations (e.g. local state change vs. hitting refresh and init state from nth creating state flickering). Let me know if I am not clear on this point! 😬

so 1 popular use case might be sth similar to sign-in with ethereum:
- user click connect wallet
- trigger a popup for user to create signature with metamask
- send that signature to backend api to login
- if success, keep the wallet connected. Otherwise, disconnect the wallet.

To my knowledge, SIWE is possibly on the roadmap. But I just think this would be helpful for other use cases beside this one as well! if there is a better way to do this, please let me know. 😄

basememara:
thx @samyouxyz for creating this. it is so much better than trying to mess around with wagmi's hooks. tbh I didn't know what wagmi was before using RainbowKit and thought it was weird I had to integrate with one of the 3rd party dependencies to do this basic functionality of the connection flow. And if one day, wagmi goes away, would be annoying to get everyone to migrate instead of abstracting it away like this.

peduarte:
we're considering this internally already, but I wanna drop some feedback. 

i've been building a few side projects with rainbowkit recently to get a better feel for it as a user. it also helps me understand parts of the DX that can be improved. i'm starting to think this callback could come in handy.

it's true that we can always check the status of the connections via `wagmi`'s `useAccount`/`useConnect` hooks. But:
1. Hooks don't re-render, so in order to conditionally render things based on connection, we still need to have a `useEffect` and set an internal state, based on the hook's `data`
2. It means, for many users, a slightly steeper learning curve, if all users want to do is check if there's a wallet connected to change something in their app. I imagine most users could do fun things with this, without going too deep into web3. Eg: interacting with contract, making transactions, signing messages, etc. 


---

@samyouxyz i've been thinking about the API for a bit, and I was wondering about two things:
1. expose a bit more information to the user, beyond a `connected` boolean. eg: the wallet used for the connection (@markdalgleish brought up that these callbacks could be useful for analytics, and I agree, so knowing which wallet was used could be helpful to track certain events)
2. follow the react's "change" event convention, so instead of two independent callbacks, it'd be `onConnectChange`

```tsx
function App() {
  const [connected, setConnected] = React.useState(false);

  return ( 
    <RainbowKitProvider onConnectChange={({ isConnected, data }) => setConnected(isConnected)}>
      {connected && <div>Enter the metaverse...</div>}
      {!connected && <ConnectButton />}
    </RainbowKitProvider>
  );
}
```

I'm thinking that `onConnectChange` would return the following object (following the same convention as [wagmi's useConnect](https://wagmi.sh/docs/hooks/useConnect) hook):

```js
{
  isConnected: boolean;
  data?: {
    address: string;
    chain: {
      id: number;
      unsupported: boolean
    }
  }
}
```

samyouxyz:
hey @peduarte. that's a very great feedback! I just got time to work on the improvement. Let me know what you think.

koolamusic:
This is beautiful 😍. How soon can we get this merged. 

Is there anything left to get it across the line

site/data/docs/connect-button.mdx
```diff
@@ -74,6 +81,30 @@ Use the `label` prop to set a custom `ConnectButton` text.
...
+When wallet disconnects.
+
+```tsx
+<ConnectButton
+  onConnectChange={({ isConnected, data }) => {
+    !isConnected && console.log('gn!');
+  }}
+/>
+```
+
```
jxom, line 107: Should we add a note in this section that this callback will only get invoked after interacting with `ConnectButton`? Consumers could get the wrong assumption that the callback will be invoked for other cases such as: page refresh, users disconnecting the app from within their wallet, etc.

koolamusic: Yes, that make sense

peduarte: SGMT!

samyouxyz: done!

koolamusic:
@jxom @markdalgleish beeping you guys on this one.

samyouxyz:
waiting

peduarte:
the `onConnectChange` prop is currently on the `ConnectButton` itself. I [was thinking](https://github.com/rainbow-me/rainbowkit/pull/389#issuecomment-1134899536) maybe it should be on `RainbowKitProvider` instead, as typically there will only be one provider but there could be many connect buttons

peterferguson:
> the `onConnectChange` prop is currently on the `ConnectButton` itself. I [was thinking](https://github.com/rainbow-me/rainbowkit/pull/389#issuecomment-1134899536) maybe it should be on `RainbowKitProvider` instead, as typically there will only be one provider but there could be many connect buttons

@peduarte You may want different effects to run on different buttons though right?

For example, if you have a part of your application where when the user connects they have to verify their wallet and other parts where you don't need them to verify

peduarte:
@peterferguson im not sure tbh. typically, a user will connect/disconnect once, and regardless where they connected from, you'll probably want to run the same side effect

>  if you have a part of your application where when the user connects they have to verify their wallet and other parts where you don't need them to verify

i think this usually happens on user actions, not connection. if you want to do an actual "sign in" instead of "wallet connection" you'll want to sign a message regardless which button they clicked on

koolamusic:
@peduarte this would be problematic for us. 
Our platform issues certificates and credentials to users on one hand

To claim
- User connects their wallet arbitrarily to validate they own the wallet address.

To create
- User connects wallet and SIWE, where we also persist a session 

When they both disconnect, the cleanup is different. 
One needs a clear cache, the other just disconnects like a regular user. 


Unlike traditional authentication mechanisms, `<Connect />` in Web3 isn't the same. 

the approach you propose would require us to wrap RainbowKit provider anywhere that isn't the `App.ts` | _app.tsx` and move it to LayoutProvider, which doesn't sound like that much of a crime. 

However I believe it's called `onConnectChange` for a reason

peduarte:
My main concern is that having `onConnectChange` on the `ConnectButton` gives the false intent that they will be fired whenever there is a disconnect from anywhere. Nothing that couldn't be solved via docs, though.

@koolamusic im really curious to understand a bit more about your use case, so we can ensure we're providing the right API for our users <3  I get it that _to claim_ and _to create_ are different. Like, on OpenSea, for example, you can connect your wallet and see your NFTs, but in order to favourite an NFT, you gotta sign a message. Still, there is only ever one global connection entry point. So I dont get _"When they both disconnect"_. Could you help me understand a bit more? In my mind having on global `onConnectChange` is cleaner, and would allow you to do different clean ups if you need

Thanks

koolamusic:
Yeah, I totally get that documentation can help fix this ofc.

For my intending use-case One user has a session persisted in browser cookies when they SIWE. The other user completely relies on the default mechanism provided by RBK which I believe is something with localstorage x wagmi etc.

koolamusic:
> My main concern is that having `onConnectChange` on the `ConnectButton` gives the false intent that they will be fired whenever there is a disconnect from anywhere. Nothing that couldn't be solved via docs, though.
> 
> @koolamusic im really curious to understand a bit more about your use case, so we can ensure we're providing the right API for our users <3 I get it that _to claim_ and _to create_ are different. Like, on OpenSea, for example, you can connect your wallet and see your NFTs, but in order to favourite an NFT, you gotta sign a message. Still, there is only ever one global connection entry point. So I dont get _"When they both disconnect"_. Could you help me understand a bit more? In my mind having on global `onConnectChange` is cleaner, and would allow you to do different clean ups if you need
> 
> Thanks



> Could you help me understand a bit more? In my mind having on global `onConnectChange` is cleaner, and would allow you to do different clean ups if you need

Yeah, I totally get that documentation can help fix this ofc.

For my intending use-case One user has a session persisted in browser cookies when they SIWE. The other user completely relies on the default mechanism provided by RBK which I believe is something with localstorage x wagmi etc.

koolamusic:
> Could you help me understand a bit more? In my mind having on global `onConnectChange` is cleaner, and would allow you to do different clean ups if you need

Yeah, I totally get that documentation can help fix this ofc.

For my intending use-case One user has a session persisted in browser cookies when they SIWE. The other user completely relies on the default mechanism provided by RBK which I believe is something with localstorage x wagmi etc. 

I think leaving it within `ConnectButton` is the best way to go about it, as it takes a very functional approach to handling connection hooks. 

Another example could be with analytics, where a developer might want to add additional metadata for users who connect or disconnect from a certain location in the app.

williamlmao:
Please merge this! Totally need this.

peduarte:
hey frens, sorry this is taking a bit longer than expected. we've just been chatting with the wagmi core team as well to make sure that
1. we definitely need this API, or it's something that its better to be done wagmi's side
2. if we do add it, it's inline with their API, as it seems that there will be some changes in the `useConnect/useAccount` hooks

but i don't believe this PR should be blocking anyone, as it's currently possible to check whether a user's connected via wagmi's `useConnect` hook

if someone's feeling blocked by this, please let me know your use case, ideally with a reproduction sandbox

thank you!

peduarte:
wagmi is working on a “global status changes” to connection state. this will likely be added to the `useAccount` hook — you can see [the draft pr here](https://github.com/tmm/wagmi/pull/582)

with this change, you'll be able to get a callback whenever the connection status changes. this is a lot more accurate than relying on `isConnected` or `account.address`. this callback will be called when a connection is successful regardless of which button it came from

i think we should wait for this change, instead of adding new API to rainbowkit. we're trying to keep the API layer as slim as possible, and trying to avoid overlaps between rainbowkit and wagmi. if we think that's not enough we can re-open the discussion and work towards a different solution

how do you all feel about this?

nickbytes:
> i think we should wait for this change, instead of adding new API to rainbowkit. we're trying to keep the API layer as slim as possible, and trying to avoid overlaps between rainbowkit and wagmi. if we think that's not enough we can re-open the discussion and work towards a different solution

🫡 I like it sir. Agree that we should avoid overlaps in functionality + really like the latest wagmi changes.

samyouxyz:
> wagmi is working on a “global status changes” to connection state. this will likely be added to the `useAccount` hook — you can see [the draft pr here](https://github.com/tmm/wagmi/pull/582)
> 
> with this change, you'll be able to get a callback whenever the connection status changes. this is a lot more accurate than relying on `isConnected` or `account.address`. this callback will be called when a connection is successful regardless of which button it came from
> 
> i think we should wait for this change, instead of adding new API to rainbowkit. we're trying to keep the API layer as slim as possible, and trying to avoid overlaps between rainbowkit and wagmi. if we think that's not enough we can re-open the discussion and work towards a different solution
> 
> how do you all feel about this?

really appreciated for the follow up! guess we have to wait for the new big change. closing for now!

jxom:
super hot spicy alpha of what this is gonna look like in `wagmi`:

https://wagmi-hwz63ehib-zoo.vercel.app/docs/hooks/useAccount#onconnected-optional

samyouxyz:
> super hot spicy alpha of what this is gonna look like in `wagmi`:
> 
> https://wagmi-hwz63ehib-zoo.vercel.app/docs/hooks/useAccount#onconnected-optional

waiting

jxom:
this is out now in `wagmi@0.5`, get it while its hot 🥵

https://wagmi.sh/docs/hooks/useAccount#onconnect-optional

koolamusic:
Thanks for the hard work on this

wong2:
`onConnect` doesn't fire most of the time. See https://github.com/wagmi-dev/wagmi/issues/777

Not sure if it is an issue with rainbowkit or wagmi

mh-jsx:
I came across the same issue where I need a disconnect callback, you can simply use useAccount() hook:

```
const {isDisconnected } = useAccount();
```

But in my case, I needed this flag outside the component, and of course, I can't use the hook outside the function. For this, I just persist the state in browser storage (local storage). Whenever the page reloads it checks is account already disconnected or not.

```
const {isDisconnected } = useAccount();

 useEffect(() => {
    if (isDisconnected) {
      localStorage.setItem("__DISCONNECTED__", "true");
    }
  }, [isDisconnected]);
```

PR #390, opened 2022-05-17
chore: restrict pnpm to v6

saihaj:
Setting up the project for first time and I had `v7` globally installed and no dev scripts run. This will give useful error to someone to downgrade.


markdalgleish:
Thanks for the PR! Closing in favour of upgrading to v7 in #411.

Adding the engines field is a great idea but I unfortunately wasn't able to do it yet because Vercel seems to have a bug where they're still using v6 despite their docs saying otherwise. We might have to revisit that in the future.

PR #393, opened 2022-05-18
fix: flow dislocations in get a wallet

nicoglennon:
Fixes the broken flow we noticed on 'Get a Wallet'. Also adding instructions + learn url for Coinbase Wallet.


peduarte (approved):
fix validated ✅

PR #405, opened 2022-05-21
Replace `WagmiProvider` for `WagmiConfig`

### LINKED ISSUE #404: [bug] `<WagmiProvider />` is deprecated

### RainbowKit Version

0.1.1

### wagmi Version

0.4.0

### Current Behavior

Many sections across the docs recommend using the `WagmiProvider` within Rainbow Kit usage.
Currently, it behaves correctly, but it was just deprecated and suggested to use `WagmiConfig` instead.

[This is the deprecation notice](https://github.com/tmm/wagmi/commit/1928685bad22e00e9440739a3f65d8a9901716a4#diff-a462dffcb2bf1aa3ce0a57b49b45ef734e52fb3cb2d19b7233341dbd97701905L118)

### Expected Behavior

The deprecation warning should not appear

### Steps To Reproduce

Follow installation steps [here](https://www.rainbowkit.com/docs/installation), and you'll get the deprecation notice if you're using Typescript

https://github.com/tmm/wagmi/commit/1928685bad22e00e9440739a3f65d8a9901716a4#diff-1cee8646d2cfba37d6ce6a6e9a8d16f8caba0b99fc3a1ad0cb997ed8c7384d2eR4

### Anything else?

_No response_

ernestognw:
## Description
After submitting #404, I decided to replace all the instances of `WagmiProvider`, but some of them are just naming and some others are inside the examples folder, which is using an older `wagmi` version, so this needs some business decisions that I'm not sure if I can take.

I'm opening the PR but I'm hoping to get feedback on some questions:

Is there any plan to bump `wagmi`?
If so, the docs update make sense, and if not, we may lock the version on the `pnpm` lock file

Also, if it's going to be kept as it is, then some notes regarding that people should use `WagmiConfig` instead for `wagmi` versions >0.4.0

Fixes #404

### DIFF
--- a/packages/example/pages/_app.tsx
+++ b/packages/example/pages/_app.tsx
@@ -14,7 +14,7 @@ import {
 import type { AppProps } from 'next/app';
 import Head from 'next/head';
 import React, { useEffect, useState } from 'react';
-import { chain, createClient, WagmiProvider } from 'wagmi';
+import { chain, createClient, WagmiConfig } from 'wagmi';
 
 const alchemyId = '_gg7wSSi0KMBsdKnGVfHDueq6xMB9EkC';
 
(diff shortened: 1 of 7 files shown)


jxom:
Thanks for opening this PR! I might close this as #412 will make this redundant.

PR #408, opened 2022-05-22
fix: 'jitter' caused by scrollbar of Dialog.

alantoa:
## Description 

This is [react-remove-scroll-bar](https://github.com/theKashey/react-remove-scroll-bar/issues/31#issue-1233809865)‘s issue, the page will be offset to the left due to the `paddingRight:15px` of the body when opening the dialog, this will give the user the feeling of page jitter. 

https://user-images.githubusercontent.com/37520667/169689287-7bf80d8d-bb2f-4d7d-882d-7e1fa39dd367.mov

## How

Set body padding-right 0 when opening Dialog.

## Test 

https://user-images.githubusercontent.com/37520667/169689304-f16d440b-b37e-45d5-960b-13f62724898a.mov

### DIFF
--- a/packages/rainbowkit/src/components/Dialog/Dialog.tsx
+++ b/packages/rainbowkit/src/components/Dialog/Dialog.tsx
@@ -29,7 +29,12 @@ export function Dialog({ children, onClose, open, titleId }: DialogProps) {
       open && event.key === 'Escape' && onClose();
 
     document.addEventListener('keydown', handleEscape);
-
+    /**
+     * Todo: update react-remove-scroll
+     * react-remove-scroll-bar issue temporary solution.
+     * https://github.com/theKashey/react-remove-scroll-bar/issues/31
+     */
+    document.body.style.paddingRight = open ? '0px' : '';
     return () => document.removeEventListener('keydown', handleEscape);
   }, [open, onClose]);
 


peduarte:
hey, thanks for that!

Yeah RainbowKit uses [react-remove-scroll](https://github.com/theKashey/react-remove-scroll), which depends on [react-remove-scroll-bar](https://github.com/theKashey/react-remove-scroll-bar)

so is this something that should be fixed by the dependency? i'm just curious because we also use `react-remove-scroll` on the Radix Dialog and this issue isn't present. It sets the `padding-right: 0` on the body. Do we know when this issue was introduced?

Radix uses [2.4.0](https://github.com/radix-ui/primitives/blob/main/packages/react/dialog/package.json#L33) and RainbowKit [2.5.3](https://github.com/rainbow-me/rainbowkit/blob/main/packages/rainbowkit/package.json#L65)

alantoa:
Hey @peduarte!
this is a [sequence issue](https://github.com/theKashey/react-remove-scroll-bar/blob/2eae7247eeb133c229c7b4617a8cf61570b5af6d/src/utils.ts#L28), `padding-right` are updated when you open the Dialog, use `react-remove-scroll v2.4.0` still has this issue, I tried.

Radix did some special treatment, you can see [this](https://github.com/radix-ui/primitives/blob/6c4e20d6aed5d3a7f374c5e6474c6f94b6759529/packages/react/dialog/src/Dialog.tsx#L208), but I didn't try carefully.

I think setting the body `paddingRightn:0` first can solve the current issue, `react-remove-scroll`'s author is also currently working on this issue. ( It would be better if it could be the same as Redux)

alantoa:
thank you for looking into this!
let's wait for this.

markdalgleish:
This issue was fixed in react-remove-scroll v2.5.4, and we've just rolled out the update in RainbowKit v0.3.1.

PR #410, opened 2022-05-23
chore: include md/mdx in lint scripts

markdalgleish:
I noticed in the Steakwallet PR that md/mdx files weren't being checked in our lint script which can lead to PRs including unrelated formatting changes applied on save by their editor.


peduarte (approved):
good stuff!

PR #412, opened 2022-05-23
feat: upgrade to wagmi 0.4

jxom:
Upgrading `wagmi` to `^0.4` and refactoring the `configureChains` implementation to use [wagmi's configureChains API](https://wagmi.sh/docs/providers/configuring-chains). This change will make RainbowKit's `apiProvider` export redundant as consumers can now import providers from wagmi. 

I thought I may as well remove the `apiProvider` export instead of marking it deprecated as it will be quite a simple migration, however, if we want to hold off on breaking changes, I'm happy to deprecate the `apiProvider` export instead.


packages/rainbowkit/src/chains/configureChains.ts
```diff
@@ -1,66 +1,36 @@
...
+      targetQuorum?: number;
+      minQuorum?: never;
+    }
+  | {
+      targetQuorum: number;
+      minQuorum?: number;
+    }
+);
 
 export const configureChains = <
```
markdalgleish, line 18: We chatted about this offline — posting here for visibility.

This `configureChains` function in RainbowKit now wraps the same function from wagmi. The only reason this has been left in RainbowKit is because we have our own `Chain` type that extends wagmi's. We're going to look into whether we can make wagmi's function generic so that we don't need to provide our own wrapper.

markdalgleish (approved):
Love it! ♥️

Make sure to hold off on merging this until we're ready to cut a release so the docs don't go out early with the new API.

PR #413, opened 2022-05-23
fix: modal height moved due to scrollbox

nicoglennon:
fix this extra spacing

Screen Shot 2022-05-23 at 12 54 18 PM


peduarte (approved):
🪄

PR #427, opened 2022-05-26
chore: fix package script filters in pnpm v7

markdalgleish:
Similar to the `--recursive` flag, `--filter` now needs to appear before the script name.


nicoglennon (approved):
🏃

PR #439, opened 2022-05-27
feat: add anchor scrolling for documentation site

tianenpang:
Add anchor scrolling for documentation site.

### DIFF
--- a/site/components/MdxComponents/MdxComponents.tsx
+++ b/site/components/MdxComponents/MdxComponents.tsx
@@ -63,17 +63,7 @@ export const components = {
   ),
   h3: ({ id, ...props }) => {
     return (
-      <Box
-        id={id}
-        marginBottom="2"
-        marginTop="11"
-        scrollMarginTop={{
-          lg: 'scrollTop1',
-          md: 'scrollTop2',
-          sm: 'scrollTop3',
-          xs: 'scrollTop3',
-        }}
-      >
+      <Box id={id} marginBottom="2" marginTop="11" scrollMarginTop="12">
         <Link href={`#${id}`}>
           <Text
             as="h3"
@@ -87,17 +77,7 @@ export const components = {
     );
   },
   h4: ({ id, ...props }) => (
-    <Box
-      id={id}
-      marginBottom="2"
-      marginTop="10"
-      scrollMarginTop={{
-        lg: 'scrollTop1',
-        md: 'scrollTop2',
-        sm: 'scrollTop3',
-        xs: 'scrollTop3',
-      }}
-    >
+    <Box id={id} marginBottom="2" marginTop="10" scrollMarginTop="12">
       <Link href={`#${id}`}>
         <Text
           as="h4"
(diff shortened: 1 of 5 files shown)


site/components/MdxComponents/MdxComponents.tsx
```diff
@@ -63,27 +63,45 @@ export const components = {
...
+        </Link>
+      </Box>
     );
   },
-  h4: props => (
-    <Text
-      as="h4"
+  h4: ({ id, ...props }) => (
+    <Box
+      as="span"
```
peduarte, line 88: could we use `Box` as `div`, instead of `span`? That way we don't need to also add `display` support to the `Box`.

peduarte:
Thanks for this, nice improvement 👌

site/components/MdxComponents/MdxComponents.tsx
tianenpang: Updated with responsive scroll top space keep 12 pixels away from the header.

site/components/Text/Text.tsx
```diff
@@ -18,6 +18,7 @@ type Props = {
     | 'span';
   children?: React.ReactNode;
   color?: BoxProps['color'];
+  display?: BoxProps['display'];
```
peduarte, line 21: Adding `display` to `Box` shouldn't be needed anymore, right?

site/components/MdxComponents/MdxComponents.tsx
```diff
@@ -63,27 +63,51 @@ export const components = {
...
       marginTop="10"
-      variant="title3"
-      weight="semibold"
-      {...props}
-    />
+      scrollMarginTop={{
+        lg: 'scrollTop1',
+        md: 'scrollTop2',
+        sm: 'scrollTop3',
+        xs: 'scrollTop3',
```
peduarte, line 98: i like the idea here, but for the sake of simplicity, I think what you had before is good enough - just choose a hardcoded value that works well across mobile/desktop. you had 100px before I believe?

peduarte (commented):
some comments, thanks!

site/components/Text/Text.tsx
tianenpang: It seems to me that since the heading elements are block level, it needs to keep `display` or add `display="inline-flex"` to the `Link`?

peduarte: Oh sorry, i misread, i thought this was `Box`, but it's `Text` 😅

site/components/MdxComponents/MdxComponents.tsx
tianenpang: Have tried to be mobile/desktop compatible in hard-code way, but the overlap with `Header` is not looking nice. 

If still feel not simplicity, I can revert the old way and set to 140px (The best top space for mobile/desktop I think).

peduarte: I tried setting it to `12` ('128px') and it worked fine on both mobile and desktop, is that overlapping for you?

tianenpang: I tried to align with the text in the sidebar 🤣 

Let me setting it to `12` quikly, should i move the `scrollMarginTop` to `unresponsiveProperties`?

peduarte: i don't mind, it's up to you

peduarte (approved):
awesome! thanks for your contribution fren

PR #444, opened 2022-05-30
feat: add create-rainbowkit

markdalgleish:
I'm leaving this undocumented for now so that we can trial it out first. If everything looks good once published, we can update the docs to make this more prominent.

Usage (once it's live): `pnpm create @rainbow-me/rainbowkit` (or `yarn create` or `npm init`)

To test locally, first compile the CLI in watch mode:

```bash
pnpm dev:cli
```

Then, in another terminal, run create-rainbowkit and start a dev server against the generated app:

```
pnpm test:cli:dev
```

There's more detail provided in the contribution guide.

The generated app is checked in since it's part of the monorepo and needs to be present when running `pnpm install`. The idea is that any changes to the template should also be reflected in updates to the generated test app.

Note: I originally tried to gitignore the generated app but found this caused issues with lock file churn since pnpm would remove it from the lock file if the app wasn't present, but this would break CI. By committing the generated app, we avoid forcing everyone to generate the test app themselves all the time, instead only generating it when the template changes and then committing the result.


packages/create-rainbowkit/generated-test-app/pages/_app.tsx
```diff
@@ -0,0 +1,47 @@
...
+    chain.mainnet,
+    chain.polygon,
+    chain.optimism,
+    chain.arbitrum,
+    ...(process.env.NEXT_PUBLIC_ENABLE_TESTNETS === 'true'
+      ? [chain.goerli, chain.kovan, chain.rinkeby, chain.ropsten]
+      : []),
+  ],
+  [
+    alchemyProvider({ alchemyId: '_gg7wSSi0KMBsdKnGVfHDueq6xMB9EkC' }),
```
nickbytes, line 20: Wondering if we need to drop a comment that this is a test ID from alchemy?

nickbytes (approved):

PR #445, opened 2022-05-30
chore: add dedicated dev:example and dev:site scripts

markdalgleish:
Since the `dev` script runs both the example app and the site, and the site is a lot heavier than the example app, these new scripts allow you to target the specific project you're working on.


peduarte (approved):
🙏

PR #451, opened 2022-05-31
feat: custom avatar

nicoglennon:
add a `RainbowKitProvider` prop that takes an `AvatarComponent` to display a custom avatar.

Screen Shot 2022-05-31 at 3 54 37 PM


packages/rainbowkit/src/components/RainbowKitProvider/AvatarContext.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export type AvatarComponentProps = {
+  address: string;
+  ensImage?: string | null;
+  size: number;
+};
+export type AvatarComponent = React.FunctionComponent<AvatarComponentProps>;
+
+export const defaultAvatar = {
+  component: undefined,
```
markdalgleish, line 11: To reduce code paths, could we make an EmojiAvatar component and use it as the default value? This way we can always just render the AvatarComponent from context, even when it's our own.

packages/rainbowkit/src/components/RainbowKitProvider/RainbowKitProvider.tsx
```diff
@@ -50,6 +50,9 @@ export interface RainbowKitProviderProps {
     disclaimer?: DisclaimerComponent;
   };
   coolMode?: boolean;
+  avatar?: {
+    component?: AvatarComponent;
```
markdalgleish, line 54: Could we flatten this to just `avatar`? e.g. `<RainbowKitProvider avatar={CustomAvatar} />`

packages/rainbowkit/src/components/Avatar/Avatar.tsx
```diff
@@ -47,7 +51,7 @@ export function Avatar({ address, imageUrl, loading, size }: AvatarProps) {
         }}
         userSelect="none"
       >
-        {imageUrl ? (
+        {imageUrl && !overrideEnsImage ? (
```
markdalgleish, line 54: Since the avatar component is passed an ENS image and can decide how to manage it, do we need to include this flag? I feel like alternating between the custom avatar and the ENS image should be managed within the avatar component and this would help reduce complexity here - happy to chat about this if I'm missing something though.

nicoglennon:
Adding the changeset and doc changes in a bit

packages/rainbowkit/src/components/RainbowKitProvider/RainbowKitProvider.tsx
nicoglennon: i would do so if we didn't have the `overrideEnsImage` prop, which you mentioned you'd want to remove below

packages/rainbowkit/src/components/Avatar/Avatar.tsx
nicoglennon: originally i did not have `overrideEnsImage` and simply did the `avatar={CustomAvatar}` pattern, but as I was building the little custom avatar for the demo it felt annoying having to "if then" the ENS image and add the custom css to make it fit when we already handle that image nicely ourselves. so my thought was that most developers would want to respect the ENS image of their users, while providing a custom "generative" component like our emojis for those who do not have an ENS image. devs can optionally pass `overrideEnsImage={true}` if they want to fully eject and handle all cases

happy to remove this and let devs handle the image themselves, just stating my original reasoning

nicoglennon:
@markdalgleish i had written a note earlier explaining my previous thoughts but i think you can disregard since i just adapted all your suggestions.

site/data/docs/custom-avatar.mdx
```diff
@@ -0,0 +1,45 @@
...
+import {
+  RainbowKitProvider,
+  AvatarComponent,
+} from '@rainbow-me/rainbowkit';
+import { generateColorFromAddress } from './utils';
+
+const CustomAvatar: AvatarComponent = ({ address, ensImage, size }) => {
+  const color = generateColorFromAddress(address);
+  return ensImage ? (
+    <img src={ensImage} width={size} height={size} />
```
markdalgleish, line 22: I think it'd be a good idea to add `style={{ borderRadius: 999 }}` to both of these so that devs get circles when they copy+paste this.

markdalgleish, line 22: I think it'd be a good idea to add `style={{ borderRadius: 999 }}` to both of these so that devs get circles when they copy+paste this.

markdalgleish (approved):
Just a couple of minor comments, but this looks great 👍

site/data/docs/custom-avatar.mdx
```diff
@@ -0,0 +1,45 @@
+---
+title: Custom Avatar
```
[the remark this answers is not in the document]
nickbytes: Updated

site/data/docs/custom-avatar.mdx
nickbytes: Added, this is now in `custom-avatars.mdx`

nickbytes: Resolving this since seems like the same issue?

markdalgleish (approved):

PR #479, opened 2022-06-07
fix: add WalletConnect deeplink choice to local storage

markdalgleish:
tl;dr we now set the `WALLETCONNECT_DEEPLINK_CHOICE` local storage value that WalletConnect sets in their official modal.

### DIFF
--- a/packages/example/pages/index.tsx
+++ b/packages/example/pages/index.tsx
@@ -243,9 +243,17 @@ const Example = () => {
                 <div>Transaction: {JSON.stringify(transactionData)}</div>
               )}
               {transactionError && <div>Error sending transaction</div>}
-              {signingData && <div>Data Signature: {signingData}</div>}
+              {signingData && (
+                <div style={{ wordBreak: 'break-all' }}>
+                  Data Signature: {signingData}
+                </div>
+              )}
               {signingError && <div>Error signing message</div>}
-              {typedData && <div>Typed Data Signature: {typedData}</div>}
+              {typedData && (
+                <div style={{ wordBreak: 'break-all' }}>
+                  Typed Data Signature: {typedData}
+                </div>
+              )}
               {typedError && <div>Error signing typed message</div>}
             </div>
           </div>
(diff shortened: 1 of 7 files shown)


packages/rainbowkit/src/components/ConnectOptions/MobileOptions.tsx
```diff
@@ -44,10 +44,26 @@ function WalletButton({ wallet }: { wallet: WalletConnector }) {
...
+                'WALLETCONNECT_DEEPLINK_CHOICE',
+                JSON.stringify({
+                  href: mobileUri.split('?')[0],
+                  name,
+                })
+              );
+            }
+
+            if (mobileUri.startsWith('http')) {
+              window.open(mobileUri, '_blank', 'noreferrer,noopener');
```
markdalgleish, line 60: This was added in case the fallback website is presented to the user so that it doesn't take over the current tab. This mirrors the way the official WalletConnect modal works.

nickbytes:
Seems like the WalletConnect modal removes this property on disconnect, do you think we should do that as well?

markdalgleish:
@nickbytes Ah, good catch, I'll have a look into it.

packages/example/pages/index.tsx
```diff
@@ -243,9 +243,17 @@ const Example = () => {
                 <div>Transaction: {JSON.stringify(transactionData)}</div>
               )}
               {transactionError && <div>Error sending transaction</div>}
-              {signingData && <div>Data Signature: {signingData}</div>}
+              {signingData && (
+                <div style={{ wordBreak: 'break-all' }}>
```
markdalgleish, line 247: I added this so that the layout doesn't break on mobile when signing a message.

markdalgleish:
@nickbytes I've updated the PR to clear the local storage data when disconnected. Let me know what you think.

packages/example/pages/index.tsx
nickbytes: Thanks this was sloppy by me.

nickbytes (approved):
Good stuff! Another step closer to getting our deeplinking right!

PR #483, opened 2022-06-08
chore: upgrade @wagmi/core to ^0.3.7

albertfolch-redeemeum:
### Context

The latest @wagmi/core version (0.3.7) allow specifying an ensAddress when you create a custom chain. To support that, I upgraded the package to the latest version.

Thanks for reviewing this 🙏

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,7 @@
     "@typescript-eslint/parser": "^5.5.0",
     "@vanilla-extract/esbuild-plugin": "^2.0.0",
     "@vanilla-extract/vite-plugin": "^3.1.2",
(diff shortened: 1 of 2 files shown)


markdalgleish:
Hey — thanks for the PR!

This PR doesn't impact consumers as it's only changing the version used for our docs site and example apps. Since we're not currently making use of the custom `ensAddress` feature in any of our examples, I'm not sure I understand why this PR would be necessary? Sorry if I'm missing something.

albertfolch-redeemeum:
hey @markdalgleish , sorry I got confused! it had nothing to do with this repo, apologies and thanks for your time. Closing PR

PR #491, opened 2022-06-10
docs: ensure create-rainbowkit commands always get latest

markdalgleish:
These commands sometimes hang on to cached copies of create-rainbowkit that may be out of date, so this is a good way to ensure that the cache is never used when scaffolding a new app.


nickbytes (approved):
🏎

PR #471, opened 2022-06-06
refactor: remove deprecated usage of wagmi connectors

jxom:
`wagmi@^0.4` has deprecated the usage of passing a function to `connectors` in `createClient`. Now, it is recommended to just pass an array to `connectors`. [See here](https://wagmi.sh/docs/migrating-to-03#passing-a-function-to-createclient-connectors-has-now-been-deprecated). 

~This PR just changes the return type of `connectors` in `connectorsForWallets` from a function to an array. Everything should still work as intended.~

PR #492, opened 2022-06-10
fix: ensure large lists of WalletConnect wallets work in Brave

markdalgleish:
Brave’s fingerprint prevention logic silently blocks WebSocket connections if too many are opened in the same session. Since we create a fresh WalletConnect connector instance for each wallet, consumers that have configured a large number of wallets can inadvertently break the connection flow in Brave.

To fix this, we now share WalletConnect connector instances between wallets when the connectors are being provided with the same options.


packages/rainbowkit/src/index.ts
```diff
@@ -1,6 +1,7 @@
 export * from './components';
 export { wallet } from './wallets/walletConnectors';
 export { getDefaultWallets } from './wallets/getDefaultWallets';
+export { getWalletConnectConnector } from './wallets/getWalletConnectConnector';
```
markdalgleish, line 4: This is exposed so consumers can create custom WalletConnect wallets while sharing the same connector instance. I've updated the docs to reflect this.

packages/rainbowkit/src/wallets/connectorsForWallets.ts
```diff
@@ -1,32 +1,29 @@
...
-            }". Each wallet must have its own connector instance.`
-          );
-        }
-
         let walletConnectModalConnector: Connector | undefined;
-        if (walletMeta.id === 'walletConnect' && connectionMethods.qrCode) {
+        if (
+          walletMeta.id === 'walletConnect' &&
+          connectionMethods.qrCode &&
+          !isMobile()
```
markdalgleish, line 25: This is an added optimisation. If the user is on mobile, this connector was never used, so we can avoid another websocket connection.

packages/rainbowkit/src/wallets/getWalletConnectConnector.ts
```diff
@@ -0,0 +1,35 @@
...
+  chains,
+  qrcode = false,
+}: {
+  chains: Chain[];
+  qrcode?: boolean;
+}) {
+  const options: WalletConnectConnectorOptions = {
+    chains,
+    options: {
+      qrcode,
```
markdalgleish, line 27: Note that we no longer set the `rpc` option. This is redundant since wagmi handles this internally now via `configureChains`, introduced in v0.4 which is our current peer dep version range. The RPC information is baked into the `chains` option which is also passed in here.

nickbytes (approved):
> we now share WalletConnect connector instances between wallets when the connectors are being provided with the same options.

This is great. Nice work, just tested in a couple browsers, looking good.

packages/rainbowkit/src/wallets/getWalletConnectConnector.ts
markdalgleish: I decided I'm going to roll this change back and apply it to https://github.com/rainbow-me/rainbowkit/pull/471 instead since it's related to that work. We can release this PR first as a patch before shipping a minor update.

PR #494, opened 2022-06-10
fix: handle account balance abbreviation

estrattonbailey:
Large account balances were previously shown in exponential notation. This PR abbreviates large numbers using US-standard k/m/b "units", and trims to one decimal place unless it's a zero. Balances under 10k not abbreviated and rounded to one decimal place. Values under one are rounded to three decimal places.

Considered a [native approach](https://gist.github.com/estrattonbailey/e0913ca628a37a230b5b3e12e8b4baf5) but decided against for browser compat reasons. We could potentially move this direction after sufficient testing.


packages/rainbowkit/src/components/ConnectButton/abbreviateNumber.test.ts
```diff
@@ -0,0 +1,55 @@
...
+const B = 1000 * M;
+
+describe('abbreviateNumber', () => {
+  it('passes through small numbers', () => {
+    expect(abbreviateNumber(1)).toEqual('1');
+    expect(abbreviateNumber(12)).toEqual('12');
+    expect(abbreviateNumber(123)).toEqual('123');
+    expect(abbreviateNumber(1234)).toEqual('1,234');
+    expect(abbreviateNumber(9999)).toEqual('9,999');
+    expect(abbreviateNumber(9999.1)).toEqual('9,999.1');
```
estrattonbailey, line 15: Gets a little long here — we cool with this?

packages/rainbowkit/src/components/ConnectButton/abbreviateNumber.ts
```diff
@@ -0,0 +1,35 @@
+/**
+ * Adapted from https://github.com/domharrington/js-number-abbreviate
+ */
+const units = ['k', 'm', 'b', 't'];
+
+export function abbreviateNumber(number: number, decimals: number = 1): string {
+  if (number < 10 ** 4)
+    return new Intl.NumberFormat()
+      .format(parseFloat(number.toFixed(decimals)))
+      .replace(/\.0$/, '');
```
estrattonbailey, line 10: Handles the <10k case (commas) as well as trims trailing zeros.

packages/rainbowkit/src/components/ConnectButton/abbreviateNumber.test.ts
nickbytes: Yea i think that's fine

nickbytes:
We talked about this offline, but if it's below one, like `0.516993575869330778`, it should show ~4 significant digits.

peduarte (approved):
*chefs kiss*

packages/rainbowkit/src/components/ConnectButton/abbreviateETHBalance.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  let result = String(number);
+
+  for (let i = units.length - 1; i >= 0; i--) {
+    const size = 10 ** ((i + 1) * 3);
+
+    if (size <= number) {
+      number = Math.round((number * decimalsDivisor) / size) / decimalsDivisor;
+
+      // round up
```
nickbytes, line 24: We talked about offline (posting here for others to see), but rounding here is my biggest concern. 

- if you have `0.5999`, do you really want to see that you have `0.6` vs `0.599`? 
  - I would say you want to see this "rounded down", because it's misleading and potentially upsetting to see that you don't have as much money as you hoped (expectations not met), vs potentially being surprised that you have a minuscule amount _more_. 
  - I'm trying to "think like a bank" here, since users may be using this in similar style applications.

nickbytes (changes requested):
Requesting changes for rounding. Happy to discuss more.

packages/rainbowkit/src/components/ConnectButton/abbreviateETHBalance.test.ts
```diff
@@ -0,0 +1,64 @@
...
+  });
+
+  it('abbreviates past 1b', () => {
+    expect(abbreviateETHBalance(1 * B)).toEqual('1b');
+    expect(abbreviateETHBalance(1.23 * B)).toEqual('1.2b');
+    expect(abbreviateETHBalance(10 * B)).toEqual('10b');
+    expect(abbreviateETHBalance(12.3 * B)).toEqual('12.3b');
+    expect(abbreviateETHBalance(123.4 * B)).toEqual('123.4b');
+    expect(abbreviateETHBalance(999.99 * B)).toEqual('999.9b'); // no rounds
+  });
```
estrattonbailey, line 63: @mikearndt how do ya feel about these rules? Trying to think about what level of precision is helpful at different amounts.

nickbytes (approved):

PR #496, opened 2022-06-13
fix: use mainnet ENS for L2s

nicoglennon:
christian noticed it was not ideal that when switching to an L2 your ENS information would go away. small fix to always pull ENS from mainnet throughout RK

Screen Shot 2022-06-13 at 11 24 50 AM

### DIFF
--- a/packages/rainbowkit/src/components/ConnectButton/ConnectButtonRenderer.tsx
+++ b/packages/rainbowkit/src/components/ConnectButton/ConnectButtonRenderer.tsx
@@ -30,6 +30,7 @@ import {
   useRainbowKitChainsById,
 } from '../RainbowKitProvider/RainbowKitChainContext';
 import { ShowRecentTransactionsContext } from '../RainbowKitProvider/ShowRecentTransactionsContext';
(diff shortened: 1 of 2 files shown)


nickbytes:
Is there any argument _against_ this @nicoglennon ?

packages/rainbowkit/src/components/ConnectButton/ConnectButtonRenderer.tsx
```diff
@@ -80,9 +81,13 @@ export function ConnectButtonRenderer({
 
   const { data: ensAvatar } = useEnsAvatar({
     addressOrName: accountData?.address,
+    chainId: chainMetadataByName.mainnet?.chainId,
```
nickbytes, line 84: Nit: you could just hardcode `1` here and not have to export/import other component... 🤔

holic:
This will resolve this issue, yeah? https://github.com/rainbow-me/rainbowkit/discussions/429

estrattonbailey (approved):

nicoglennon:
> This will resolve [this issue](https://github.com/rainbow-me/rainbowkit/discussions/429), yeah?

nope that is different, although i believe it only happens in dev mode (next)

PR #505, opened 2022-06-15
chore: update react-remove-scroll to v2.5.4

markdalgleish:
This update fixes an issue where page content can jump around when the modal is opened if scrollbars are permanently visible. More detail: https://github.com/theKashey/react-remove-scroll/issues/71

Closes #408.


peduarte (approved):
nicceeeee

PR #511, opened 2022-06-17
fix: automatically connect to first configured chain

markdalgleish:
No more "Wrong network" after connecting 🎉

wagmi now accepts a `chainId` option on its `useConnect` hook. Internally it will either connect to the specified chain if supported, or it will immediately switch chains after connecting if it's different to the specified chain.

### DIFF
--- a/packages/rainbowkit/src/wallets/useWalletConnectors.ts
+++ b/packages/rainbowkit/src/wallets/useWalletConnectors.ts
@@ -2,6 +2,7 @@ import { Connector, useConnect } from 'wagmi';
 import { flatten } from '../utils/flatten';
 import { indexBy } from '../utils/indexBy';
 import { isNotNullish } from '../utils/isNotNullish';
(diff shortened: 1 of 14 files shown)


packages/rainbowkit/src/wallets/useWalletConnectors.ts
```diff
@@ -14,7 +15,10 @@ export interface WalletConnector extends WalletInstance {
 }
 
 export function useWalletConnectors(): WalletConnector[] {
-  const { connectAsync, connectors: defaultConnectors } = useConnect();
+  const chains = useRainbowKitChains();
+  const { connectAsync, connectors: defaultConnectors } = useConnect({
+    chainId: chains[0]?.id,
```
nickbytes, line 20: everything is truly called _chains_ lol

nickbytes (approved):
This is an awesome quality of life improvement

PR #515, opened 2022-06-20
fix: support WalletConnect Android deep linking after connecting

### LINKED ISSUE #387: [bug] signMessage Does Not Launch Rainbow/Metamask on Android

### RainbowKit Version

0.1.0

### wagmi Version

0.3.5

### Current Behavior

When calling `signMessage` within the Chrome browser on an Android device, there is haptic feedback, but the wallet app (e.g. Rainbow, MetaMask) is not launched / brought to the foreground, requiring that the user manually open the wallet app to find the sign message prompt.

### Expected Behavior

I would expect the connected wallet (e.g. Rainbow, MetaMask) to be launched / brought to the foreground, similar to the experience of connecting a wallet. See also https://github.com/rainbow-me/rainbowkit/issues/460.

### Steps To Reproduce

1. In Chrome on an Android device, navigate to https://neon-griffin-f8c8b4.netlify.app/
2. Connect to Rainbow or MetaMask.
3. Press "Sign Message" and observe that while there is haptic feedback, the wallet app is not brought to the foreground.

https://github.com/nickcherry/rainbowkit-test

### Anything else?

_No response_

markdalgleish:
This PR adds detection for `wc:` URIs when usng WalletConnect-based wallets on Android so that deep linking works when performing wallet interactions after connecting.

Fixes #387

### DIFF
--- a/.changeset/spotty-weeks-impress.md
+++ b/.changeset/spotty-weeks-impress.md
@@ -1,0 +1,5 @@
+---
+'@rainbow-me/rainbowkit': patch
+---
+
+Support deep linking of wallet interactions for WalletConnect on Android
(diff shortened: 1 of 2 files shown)


peduarte (approved):
nice

PR #527, opened 2022-06-22
fix: update connect button styles to fix layout shift

marco:
Currently, the `ConnectButton`'s height in the disconnected state is about 1px shorter than the connected state's height (which is always fixed at 40px). This PR slightly modifies the connect button's styles to match the 40px height and fix the layout shift.

**Before**

Connecting and disconnecting shifts the text below:

before-button

**After**

Set at 40px:

after-button

### DIFF
--- a/.changeset/swift-houses-float.md
+++ b/.changeset/swift-houses-float.md
@@ -1,0 +1,5 @@
+---
+'@rainbow-me/rainbowkit': patch
+---
+
+Update connect button height to be consistent between states.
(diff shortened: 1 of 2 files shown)


markdalgleish:
Nice catch! 🙏

This will need a changeset so it goes out in a release. If you run `pnpm changeset` and follow the prompts, this should go out as a patch release. You can look at previous PRs if you need a reference for what the generated file should look like.

markdalgleish (approved):

PR #540, opened 2022-06-27
fix: ensure blank tab isn't left behind after connecting on iOS

### LINKED ISSUE #524: [bug] Connecting to WalletConnect-based wallets in Chrome on iOS returns users to a blank tab

### RainbowKit Version

0.3.4

### wagmi Version

0.4.10

### Current Behavior

When connecting to a WalletConnect-based wallet in Chrome on iOS, once you've connected and been returned to the browser, you're presented with an empty tab at `about:blank`. You then need to manually close to return to the original page that triggered the connection.

### Expected Behavior

You should go back to the same page you were on previously.

### Steps To Reproduce

1) Go to https://rainbowkit-example.vercel.app in Chrome on iOS.
2) Tap "Connect Wallet".
3) Tap "Rainbow" -> Goes to Rainbow app.
4) Tap "Connect" -> Goes back to browser.
5) See `about:blank` tab.
6) Close tab to return to original page.

https://rainbowkit-example.vercel.app

### Anything else?

It's likely that this affects other browsers, but so far I've only attempted to reproduce this in Chrome.

markdalgleish:
Fixes #524.

Using `window.open` causes issues on iOS in non-Safari browsers and WebViews where a blank tab is left behind after connecting. This is especially bad in some WebView scenarios (e.g. following a link from Twitter) where the user doesn't have any mechanism for closing the blank tab.

For whatever reason, links with a target of "_blank" don't suffer from this problem, and programmatically clicking a detached link element with the same attributes also avoids the issue.

I've reproduced this issue in the following iOS contexts and confirmed that this fixes it:
- Chrome
- Brave
- Firefox
- Twitter WebView (I sent a DM to myself to test)

I've also confirmed that Android still works as expected using Chrome, Brave and Twitter in-app browser.

Here's an example before/after in Chrome on iOS:

Before:

https://user-images.githubusercontent.com/696693/175839319-78b22ec6-9dba-44d4-bedf-a80a9068ffe5.mp4

After:

https://user-images.githubusercontent.com/696693/175839327-56fec81d-480b-42bb-a360-c25fece1b197.mp4

### DIFF
--- a/.changeset/shaggy-doors-grab.md
+++ b/.changeset/shaggy-doors-grab.md
@@ -1,0 +1,5 @@
+---
+'@rainbow-me/rainbowkit': patch
+---
+
+Fix issue on iOS in non-Safari browsers and WebViews where a blank tab is left behind after connecting via WalletConnect
--- a/packages/rainbowkit/src/components/ConnectOptions/MobileOptions.tsx
+++ b/packages/rainbowkit/src/components/ConnectOptions/MobileOptions.tsx
@@ -61,7 +61,7 @@ function WalletButton({ wallet }: { wallet: WalletConnector }) {
               // Workaround for https://github.com/rainbow-me/rainbowkit/issues/524.
               // Using 'window.open' causes issues on iOS in non-Safari browsers and
               // WebViews where a blank tab is left behind after connecting.
-              // This is especially bad in some WebView scenarios (i.e. following a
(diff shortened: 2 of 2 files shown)


intergalacticspacehighway (approved):
Working. Thank you, ser! 🙏

estrattonbailey (approved):

PR #474, opened 2022-06-06
feat: add torus connector

OlremProj:
There is my first try to add torus connector on rainbowkit 
So for the moment I adapt my idea from other connector already develop on the lib (metamask) 

Torus not present today on wagmi project so i create class in torus.ts to handle that tel me if it's ok for you and what part i need to correct to be more clean.
[Discuss of the idea to add torus connector link](https://github.com/rainbow-me/rainbowkit/discussions/459)

Last commit resolve all my problem, last question is for the moment torus start login on the first chain set in the ui because the lib didn't handle change chain so to handle this we need to logout and reconnect with another chain if you didn't have other comment on the actual PR I add the function to my PR quickly

PR #543, opened 2022-06-27
feat: add Torus support

estrattonbailey:
A (probably incomplete) list of remaining items:
- can we hide the Torus chatbot?
- how do we like the `shouldCloseModalOnConnecting` proposed here by #474
- opening, closing, then opening the Web3Auth modal fails with `ConnectorAlreadyConnectedError: Connector already connected`
- connecting, closing the tab, opening a new tab at same URL flashes connection state then disconnects
- switching to Polygon from mainnet, back to mainnet causes disconnect, but upon refresh you're still connected
- refreshing while connected to Polygon crashes page with `network does not support ENS`
- send transaction / sign message / etc, not working


flux0uz:
Hey @estrattonbailey, any updates on this PR ?

h0tw4t3r:
@nickbytes why is this pr closed? can we get the torus back?

jacobmakarsky:
why it gone

PR #544, opened 2022-06-27
feat: add disconnect to switch network modal

nicoglennon:
Screen Shot 2022-06-28 at 2 40 36 AM


packages/rainbowkit/src/components/ConnectOptions/ConnectDetails.tsx
```diff
@@ -429,6 +429,7 @@ export function InstructionDetail({
               height="48"
               minWidth="48"
               overflow="hidden"
+              position="relative"
```
markdalgleish, line 432: Is this an accidental change?

markdalgleish (approved):
One minor note about what looks like an accidental change, but otherwise this looks great 🎉🎉

packages/rainbowkit/src/components/ConnectOptions/ConnectDetails.tsx
nicoglennon: not accidental, this fixed a safari bug i noticed with the instruction icons:
Screen Shot 2022-06-29 at 1 22 21 AM

PR #547, opened 2022-06-28
fix: store WalletConnect deep link choice for custom URL schemes

markdalgleish:
Instead of trying to detect WalletConnect URLs by inspecting the string in isolation, we now check that the wallet's connector ID is `"walletConnect"`. This fixes an issue where custom URL schemes (e.g. `imtokenv2://wc?uri=...`) weren't being detected correctly.


peduarte (approved):
killing it

PR #550, opened 2022-06-28
feat: add TokenPocket wallet

chendatony31:


nickbytes:
Thanks for the contribution here @chendatony31 – we're currently discussing out strategy for adding new wallets to the repo and how we can best serve the RainbowKit community while also not being gatekeepers/maintainers of _every_ wallet integration. I'll let you know when we have an update! Thanks for your patience!

markdalgleish:
Sorry for the delay on this, and thanks for the PR!

We've just re-worked RainbowKit internals in v0.7 so that wallets can be imported individually, ensuring our bundle size doesn't keep increasing as we continue to add more wallets.

Thanks for giving us access to your fork — I've pushed some updates to your PR migrating it to the new API. Once I've prepared a few more wallet PRs to batch together, I'll get this one merged.

markdalgleish:
I just tried this integration and noticed the following:
- On Android, tapping the TokenPocket icon opens the app but then nothing happens. At this point the app regularly crashes on me too. Have you been able to get this to work?
- On iOS, after connecting I'm not prompted to go back to my browser (or redirected automatically). Not a bug, but just flagging that there's a bit of friction there.
- On desktop, I'm not presented with a WalletConnect QR code even though the mobile apps support this. If I select WalletConnect on desktop instead of TokenPocket, I am able to connect via QR code. I recommend looking at the other WalletConnect-based wallets to get an idea of how this works. We also support presenting users with download links for both the browser extension and the mobile app to ensure that they can still get the browser extension even though they're presented with the QR code.

chendatony31:
@markdalgleish  Thank you very much for these good suggestions, we have fix these questions except the second one, which need some time for our wallet to fix it.

DanielSinclair:
@chendatony31 Circling back on this: are the issues resolved? Would be happy to look at reviewing again

PR #574, opened 2022-07-08
chore: Adds codeQL to CI

nickbytes:
- adds codeQL for javascript and typescript


estrattonbailey (approved):
LGTM, same as [the app](https://github.com/rainbow-me/rainbow/blob/develop/.github/workflows/codeqml.yml)

PR #578, opened 2022-07-11
ci: remove `typescript` from CodeQL config

estrattonbailey:
Unnecessary, `javascript` also processes our TypeScript code. See slack convo and [related app PR](https://github.com/rainbow-me/rainbow/pull/3675).

Sounds like we can leave the cron too, which could catch updates to dependencies that happen in between PRs.


estrattonbailey:
Duplicates #579

PR #583, opened 2022-07-13
fix: update codesandbox links

nicoglennon:
new codesandbox examples at https://codesandbox.io/u/rainbowdotme 🛹

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,9 +32,9 @@ For full documentation, visit [rainbowkit.com](https://rainbowkit.com).
 
 You can use the CodeSandbox links below try out RainbowKit:
 
-- with [Create React App](https://codesandbox.io/s/dn3rho)
-- with [Create React App (TypeScript)](https://codesandbox.io/s/ilfuoy)
-- with [Next.js](https://codesandbox.io/s/tmxcc0)
+- with [Create React App](https://codesandbox.io/s/rainbowkit-nextjs-gz890p)
+- with [Create React App (TypeScript)](https://codesandbox.io/s/rainbowkit-create-typescript-app-xuxnqy)
+- with [Next.js](https://codesandbox.io/s/rainbowkit-nextjs-gz890p)
 
 ## Examples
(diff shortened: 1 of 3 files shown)


peduarte (commented):
niiice! 

should we update `installation.mdx` as well? it also references some of the old urls 

also might be worth double checking that the url in `introduction.mdx` is still correct

README.md
```diff
@@ -32,9 +32,9 @@ For full documentation, visit [rainbowkit.com](https://rainbowkit.com).
 
 You can use the CodeSandbox links below try out RainbowKit:
 
-- with [Create React App](https://codesandbox.io/s/dn3rho)
-- with [Create React App (TypeScript)](https://codesandbox.io/s/ilfuoy)
-- with [Next.js](https://codesandbox.io/s/tmxcc0)
+- with [Create React App](https://codesandbox.io/s/rainbowkit-nextjs-gz890p)
```
markdalgleish, line 35: This CRA link is pointing to the Next.js sandbox.

README.md
```diff
@@ -32,9 +32,9 @@ For full documentation, visit [rainbowkit.com](https://rainbowkit.com).
...
+- with [Create React App (TypeScript)](https://codesandbox.io/s/rainbowkit-create-typescript-app-xuxnqy)
```
markdalgleish, line 36: I tried using the app in this sandbox and I get an error after trying to connect saying `No wagmi client found`. Is this a new issue? My guess is it's an ESM/CJS issue where multiple copies of wagmi are being loaded so the provider and hooks don't line up.

nicoglennon: the issues i was seeing the other day with ESM just disappeared when i revisited the sandbox later that night... not sure what is going on but seems unstable

nickbytes (approved):

README.md
nickbytes: Not seeing any issues w/ the above two

PR #585, opened 2022-07-14
feat: add initialChain prop to RainbowKitProvider

### LINKED ISSUE #563: [bug] "automatically connect to first configured chain" keeps invoking annoying network switch
### RainbowKit Version

0.4.0

### wagmi Version

0.5.5

### Current Behavior

In the app I have, the network can be chosen before connecting the wallet, i.e. the user can interact with the app and leave the wallet connection for last. When user finally connects then a switch to the first chain is always triggered which can be completely off from what was chosen in the app. When the user doesn't abide to the network switch then **the wallet is not connected**.

In a nutshell, `wagmi` makes it possible to pass in the Chain ID when using `useConnect` _but_ RainbowKit hides this and instead forces a switch to the arbitrary first network.

### Expected Behavior

* Not hide the features that wagmi exposes -- make it possible to pass in Chain ID when connecting with RainbowKit
* Not _force_ a network switch at all when connecting with RainbowKit

### Anything else?

Issue introduced in this PR: https://github.com/rainbow-me/rainbowkit/pull/511

markdalgleish:
RainbowKit (as of v0.3.2) automatically connects to the first chain in the `chains` array passed to `RainbowKitProvider`. This behavior can now be customized via the `initialChain` prop.

The initial chain can be configured using a chain ID.

```tsx
<RainbowKitProvider chains={chains} initialChain={1}>
```

As a convenience, you can also pass a chain object.

```tsx
<RainbowKitProvider chains={chains} initialChain={chain.mainnet}>
```

Fixes #563.

### DIFF
--- a/.changeset/dry-goats-pay.md
+++ b/.changeset/dry-goats-pay.md
@@ -1,0 +1,16 @@
+---
+'@rainbow-me/rainbowkit': minor
+---
+
+**Breaking:** Added new required `initialChain` prop to `RainbowKitProvider`
+
+RainbowKit previously (as of v0.3.2) automatically connected to the first chain in the `chains` prop on `RainbowKitProvider`. This behavior no longer provided by default. Instead, you must now explicitly specify which chain you want users to initially connect to.
+
+**Migration guide**
+
+If you want to keep the existing behavior, set the `initialChain` prop to `chains[0]`.
+
+```diff
+-<RainbowKitProvider chains={chains}>
++<RainbowKitProvider chains={chains} initialChain={chains[0]}>
+```
--- a/.changeset/dry-goats-pay.md
+++ b/.changeset/dry-goats-pay.md
@@ -1,16 +1,19 @@
 ---
-'@rainbow-me/rainbowkit': minor
+'@rainbow-me/rainbowkit': patch
 ---
 
-**Breaking:** Added new required `initialChain` prop to `RainbowKitProvider`
+Added `initialChain` prop to `RainbowKitProvider`
 
-RainbowKit previously (as of v0.3.2) automatically connected to the first chain in the `chains` prop on `RainbowKitProvider`. This behavior no longer provided by default. Instead, you must now explicitly specify which chain you want users to initially connect to.
+RainbowKit (as of v0.3.2) automatically connects to the first chain in the `chains` array passed to `RainbowKitProvider`. This behavior can now be customized via the `initialChain` prop.
 
-**Migration guide**
+The initial chain can be configured using a chain ID.
 
-If you want to keep the existing behavior, set the `initialChain` prop to `chains[0]`.
+```tsx
+<RainbowKitProvider chains={chains} initialChain={1}>
+```
 
-```diff
--<RainbowKitProvider chains={chains}>
-+<RainbowKitProvider chains={chains} initialChain={chains[0]}>
+As a convenience, you can also pass a chain object.
+
+```tsx
+<RainbowKitProvider chains={chains} initialChain={chain.mainnet}>
 ```
--- a/examples/with-create-react-app/src/index.tsx
+++ b/examples/with-create-react-app/src/index.tsx
@@ -45,7 +45,7 @@ const root = ReactDOM.createRoot(
 root.render(
   <React.StrictMode>
     <WagmiConfig client={wagmiClient}>
-      <RainbowKitProvider chains={chains}>
+      <RainbowKitProvider chains={chains} initialChain={chains[0]}>
         <App />
       </RainbowKitProvider>
     </WagmiConfig>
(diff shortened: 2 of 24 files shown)


saihaj (commented):
Thoughts on using an object to represent chains:
**Proposal 1**
```js
const chains = {
  1: {
    rpcUrls: {
      default: 'https://mainnet.infura.io/',
    },
  },
  4: {
    rpcUrls: {
      default: 'https://rinkeby.infura.io/',
    },
  },
  1337: {
    rpcUrls: {
      default: 'http://localhost:8545',
    },
  },
};
```

So now you can do

```diff
-<RainbowKitProvider chains={chains}>
+<RainbowKitProvider chains={chains} initialChain={1}>
```

**Proposal 2**  inspired from Stitches [`defaultVariant` style](https://stitches.dev/docs/variants#default-variants)
```js
const chains = {
  1: {
    rpcUrls: {
      default: 'https://mainnet.infura.io/',
    },
  },
  4: {
    rpcUrls: {
      default: 'https://rinkeby.infura.io/',
    },
  },
  1337: {
    rpcUrls: {
      default: 'http://localhost:8545',
    },
  },
  default: {
    id: 1,
  },
};
```

So now you can just pass as is

```tsx
<RainbowKitProvider chains={chains}>
```

markdalgleish:
@saihaj If you check out the linked issue (https://github.com/rainbow-me/rainbowkit/issues/563) you can see they're wanting to dynamically set the initial chain based on app state that the user controls before connecting, whereas the chain config for wagmi/RainbowKit is designed to be static.

saihaj:
> @saihaj If you check out the linked issue (#563) you can see they're wanting to dynamically set the initial chain based on app state that the user controls before connecting, whereas the chain config for wagmi/RainbowKit is designed to be static.

yes with what I am recommending is somewhat nicer API to define the initial chain that user should connect to. The config can define the `default`/`initial` chain it should connect with.

kasparkallas:
@markdalgleish This PR would fix the issue we're having and I would _love_ to see this merged!

Some comments:
* Documentation & migration guide & examples should mention/show that you can pass in either the chain type or the chain ID (as I realized from the code).
* I could imagine `openConnectModal` also taking the `initialChain` as input (to override the context's one) _but_ this is more about general API design then an actual problem I'm having.

@saihaj I prefer the version with `initialChain` passed in separate from the `chains` list.
The proposal 2 example you provider would break the simple and logical structure: 
```ts
interface Chains { 
    [chainId]: Chain 
}
```
The proposal 2 is also not equivalent to the `defaultVariants` from Stitches because the structure there is:
```ts
const styles = {
    variants: { ... },
    defaultVariants: { ... }
}
```
For it to be equivalent then the RainbowKit's object would have to have this structure:
```ts
const chainOptions = {
    chains: [ ... ],
    initialChain: ...
}
```

markdalgleish:
Based on an offline conversation with @peduarte, I've changed this prop from required to optional to reduce boilerplate, particularly for consumers who only have a single chain configured. We may revisit this in the future, but making it optional is an easy no-regrets move.

@kasparkallas Thanks for the feedback. I've updated the docs to highlight that the prop accepts an ID as well as a chain object.

kasparkallas:
> Based on an offline conversation with @peduarte, I've changed this prop from required to optional to reduce boilerplate, particularly for consumers who only have a single chain configured. We may revisit this in the future, but making it optional is an easy no-regrets move.

My biggest issue with this is that `wagmi` provides an option to connect to a wallet without a network switch triggering but it is now removed from RainbowKit. A customizability option -- gone.

markdalgleish:
@kasparkallas We're trying to avoid immediately presenting users with the "Wrong network" state after connecting. Is there a reason I'm missing that you'd want this behaviour?

kasparkallas:
> @kasparkallas We're trying to avoid immediately presenting users with the "Wrong network" state after connecting.

I understand and I didn't recommend against the chosen default behaviour -- I'm was stating a fact that the customizability feature was removed.

> Is there a reason I'm missing that you'd want this behaviour?

After this PR get merged and I can pass in the chain ID then I will personally not have an immediate reason for this behaviour.

But I'm recommending a rule of thumb not to remove or hide wagmi's features if there's not a really good reason to do so. It's better API design in my opinion (you can disagree). 

Just allow the user not to connect to any network _because_ it's part of wagmi's API / feature set and there's no technical reason to hide it. Keep RainbowKit a superset of wagmi, not a subset.

I hope I'm not coming off a condescending. Really looking forward to seeing this PR merged and thanks for the awesome package!

nickbytes (approved):

stefan-girlich:
I am in a similar situation as @kasparkallas (see [comment](https://github.com/rainbow-me/rainbowkit/pull/585#issuecomment-1188762757) and #563 ):
I want to allow the user to stay on their currently selected network (e.g. via MetaMask) as long as the network is contained in `chains`. This is critical for my application since SIWE is used and switching networks would invalidate the server-side option.
Using `initialChain` does not seem feasible since I cannot determine the network selected in the users's wallet before they connect it. @kasparkallas @markdalgleish - how could I allow the user to connect to any network in `chains` (as it was before https://github.com/rainbow-me/rainbowkit/pull/511 ) ? Thanks a lot!

PR #594, opened 2022-07-16
Add useOpenConnectModal

malonehedges:
This adds a hook `useOpenConnectModal` to rainbowkit by wrapping the entire application in a custom button.

I wasn't able to verify if this change works because I don't use pnpm and couldn't get the dependencies for the example folders to resolve properly, so this should be treated as a proof of concept.

I've done this in an app (without modifying RainbowKit) with much success, allowing me to not having to wrap random components in a custom button but instead use a hook and one shared modal component.


shrugs:
imo the more accurate change that should be made is that rainbow-kit internal state (specifically the state managed by `<ConnectButton />` should be part of a global context, allowing for more custom integrations and hooks like `useAccountModalState()`. the current architecture assumes a static "connect wallet" area of a website that will always exist and is the only place this context will be required. which isn't necessarily accurate the more custom and specific a web3 integration becomes

markdalgleish:
Thanks for the PR! This is something I was actually already looking into before this PR was opened, and we just shipped this (along with Hooks for opening the account and chain modals too) in [RainbowKit v0.4.3.](https://github.com/rainbow-me/rainbowkit/releases/tag/%40rainbow-me%2Frainbowkit%400.4.3)

PR #597, opened 2022-07-18
feat: add modal Hooks

markdalgleish:
This PR introduces the following Hooks:

- `useConnectModal`
- `useAccountModal`
- `useChainModal`

To accomodate this work, I made the following refactors:

- Moved rendering of the modal components from `ConnectButtonRenderer` up to `RainbowKitProvider`, which it renders via `ModalProvider`. I kept all modal-related code in a separate `ModalContext` file to keep it manageable and to avoid circular dependencies. This was needed since it's now possible to avoid `ConnectButton`/`ConnectButton.Custom` entirely.
- For the same reason as the previous point, image preloading is now triggered in `RainbowKitProvider` rather than `ConnectButtonRenderer`.
- The internal modal components now only need `open` and `onClose` props. We previously passed a lot of additional data props around to ensure we didn't fire off multiple requests for the same information, but this is no longer needed now that wagmi leverages React Query to de-dupe requests.

From a docs perspective, I added a new page called "Modal Hooks" under advanced usage. I wasn't entirely sure where it belongs so I thought it'd be worth calling that out.

Since `ConnectButton.Custom` no longer renders the modal components internally, it makes sense that this should be converted into a Hook, but I decided to leave this for a future PR.

**Example usage**

```tsx
import {
  useConnectModal,
  useAccountModal,
  useChainModal,
} from '@rainbow-me/rainbowkit';

export const YourApp = () => {
  const { openConnectModal } = useConnectModal();
  const { openAccountModal } = useAccountModal();
  const { openChainModal } = useChainModal();

  return (
    <>
      {openConnectModal && (
        <button onClick={openConnectModal} type="button">
          Open Connect Modal
        </button>
      )}
      
      {openAccountModal && (
        <button onClick={openAccountModal} type="button">
          Open Account Modal
        </button>
      )}
      
      {openChainModal && (
        <button onClick={openChainModal} type="button">
          Open Chain Modal
        </button>
      )}
    </>
  );
};
```


packages/rainbowkit/src/components/RainbowKitProvider/RainbowKitProvider.tsx
```diff
@@ -97,34 +101,36 @@ export function RainbowKitProvider({
...
-                              : null,
-                          ].join(''),
-                        }}
-                      />
-                      {children}
-                    </div>
-                  ) : (
-                    children
-                  )}
+                  <ModalProvider>
```
markdalgleish, line 104: This is nested at this depth because the modal UI needs access to the other context values.

packages/rainbowkit/src/components/ConnectButton/ConnectButtonRenderer.tsx
```diff
@@ -196,30 +136,10 @@ export function ConnectButtonRenderer({
         chainModalOpen,
         connectModalOpen,
         mounted,
-        openAccountModal,
-        openChainModal,
-        openConnectModal,
+        openAccountModal: openAccountModal ?? noop,
+        openChainModal: openChainModal ?? noop,
+        openConnectModal: openConnectModal ?? noop,
```
markdalgleish, line 141: We fall back to a no-op function here to maintain backwards compatibility.

nickbytes (approved):
People are going to love this

malonehedges:
thank you!

PR #604, opened 2022-07-20
fix: avoid ENS requests when mainnet isn't configured

markdalgleish:
Related discussions:

- https://github.com/rainbow-me/rainbowkit/discussions/593
- https://github.com/rainbow-me/rainbowkit/discussions/592
- https://github.com/rainbow-me/rainbowkit/discussions/429

### DIFF
--- a/packages/rainbowkit/src/hooks/useEnsChain.ts
+++ b/packages/rainbowkit/src/hooks/useEnsChain.ts
@@ -1,0 +1,13 @@
+import { chain } from 'wagmi';
+import { useRainbowKitChains } from './../components/RainbowKitProvider/RainbowKitChainContext';
+
+export function useEnsChain() {
+  const rainbowKitChains = useRainbowKitChains();
+
+  const chainId = chain.mainnet.id;
+  const enabled = rainbowKitChains.some(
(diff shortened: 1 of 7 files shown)


packages/rainbowkit/src/hooks/useEnsChain.ts
```diff
@@ -0,0 +1,13 @@
...
+export function useEnsChain() {
+  const rainbowKitChains = useRainbowKitChains();
+
+  const chainId = chain.mainnet.id;
+  const enabled = rainbowKitChains.some(
+    rainbowKitChain => rainbowKitChain.id === chainId
+  );
+
+  return { chainId, enabled };
+}
```
peduarte, line 13: nit: since the `chainId` is hardcoded to mainnet, should this hook be named `useMainnetEnsChain`?

peduarte (approved):

packages/rainbowkit/src/hooks/useEnsChain.ts
peduarte: nice, yeah that makes more sense to me now <3

PR #623, opened 2022-07-29
docs: update `alchemyId` to `apiKey` to adhere to the new wagmi API

Dhaiwat10:
Noticed while trying to follow the installation docs that you need to provide an `apiKey` instead of an `alchemyId` in the `configureChains` method now. Updated this in your installation docs.


jxom (approved):

jxom (changes requested):
Looks good! We also need to update the "Chains" & "Custom Wallet List" docs with the new usage.

Dhaiwat10:
@jxom good catch, done

peduarte (approved):
excellent stuff frens! <3

jxom (approved):

PR #636, opened 2022-08-05
feat: add authentication and SIWE/NextAuth support

markdalgleish:
See changesets for details.

Note that I've added an "authentication" toggle to the example app so you can play around with it, but I've left it toggled off by default.

How it works:

- The connect modal checks the authentication status to see if it's "unathenticated". If it's in an unauthenticated state, the modal renders the `SignIn` component instead of closing.
- The `SignIn` component uses context to ask for an "authentication adapter" which is an object with the following methods: `getNonce`, `createMessage`, `getMessageBody`, `verify` and `signOut`. This keeps our UI generic. This adapter is provided by the `RainbowKitAuthenticationProvider` component.
- Ideally consumers don't use `RainbowKitAuthenticationProvider` directly because it's so unopinionated. It's much nicer to have a plugin-style provider for specific auth backends + message format combinations. We ship `RainbowKitSiweNextAuthProvider` as a good example of this pattern. Since this provider knows it's using SIWE and NextAuth, it handles most of the boilerplate, relying on the existence of NextAuth's `SessionProvider` and utility functions.
- On successful authentication, the auth status is updated which causes the modal to close.

EDIT: One gotcha I've found is that server-side authentication only works for Vercel preview deployments when looking at a specific commit, not a branch deployment, e.g. https://rainbowkit-example-hoa6ad4bm-rainbowdotme.vercel.app, so you'll need to ignore the Vercel bot comment if you want to test auth. This is because you can only log in from the domain that matches `process.env.VERCEL_URL`. I'm not sure if there's a way to work around this limitation.


packages/example/lib/useConnectionStatus.ts
```diff
@@ -0,0 +1,27 @@
+import { useSession } from 'next-auth/react';
```
jxom, line 1: I am worried about this hook combinoing both connection & auth state as consumers may use it as a way to retrieve the source of truth for combined connection & authentication state. These states can get very complex & I am also not sure how I feel about `!isConnected === 'disconnected'` as `!isConnected` could mean either `connecting`, `reconnecting` or `disconnected` from wagmi. 

Wonder if we can just use `useAccount` & `useSession` hook directly in the usage? If consumers want to abstract this, they can make their own `useConnectionStatus` hook like RainbowKit has done in `src/hooks`. 

```tsx
const { status } = useSession()

const ready = mounted && (authEnabled ? status !== 'loading' : true);
const connected = authEnabled ? status === 'authenticated' : isConnected;
```

Thoughts?

jxom:
Looping from internal convo:

I wonder if we should add https://next-auth.js.org/tutorials/securing-pages-and-api-routes#server-side to the example so we can remove the "flash of no connect button". 

I also wonder if we should provide a client-side only way to mitigate the flash too. Such as persisting authentication status to local storage (this is what wagmi does).

packages/rainbowkit-siwe-next-auth/src/useSiweNextAuth.ts
```diff
@@ -0,0 +1,57 @@
...
+      createAuthenticationAdapter({
+        createMessage: ({ address, chainId, nonce }) =>
+          new SiweMessage({
+            address,
+            chainId,
+            domain: window.location.host,
+            nonce,
+            statement,
+            uri: window.location.origin,
+            version: '1',
```
jxom, line 28: is this ok to be static?

packages/rainbowkit-siwe-next-auth/src/useSiweNextAuth.ts
```diff
@@ -0,0 +1,57 @@
...
+
+export function useSiweNextAuth({
+  statement = 'Sign in with Ethereum to the app.',
+}: UseSiweNextAuthOptions = {}): AuthenticationConfig<SiweMessage> {
+  const { status } = useSession();
+  const adapter = useMemo(
+    () =>
+      createAuthenticationAdapter({
+        createMessage: ({ address, chainId, nonce }) =>
+          new SiweMessage({
```
jxom, line 21: does the consumer need to pass any of these too?

https://github.com/spruceid/siwe/blob/main/packages/siwe/lib/client.ts#L41-L54

markdalgleish: Some of those values are too dynamic to be passed in via the provider. I wonder if this means we need to support passing in a function to dynamically resolve them.

e.g.

```tsx
function createMessage({ address, chainId, nonce }) {
  return new SiweMessage({
    address,
    chainId,
    nonce,
    ...etc
   });
}

<RainbowKitSiweNextAuthProvider createMessage={createMessage}>
```

Note that this means `siwe` would need to be a peer dep.

Or maybe just providing the options?

```tsx
function getMessageOptions({ address, chainId, nonce }) {
  return ({
    address,
    chainId,
    nonce,
    requestId: '...'
  });
}

<RainbowKitSiweNextAuthProvider getMessageOptions={getMessageOptions}>
```

This could even be simplified to provide defaults?

```tsx
function getMessageOptions({ defaults }) {
  return ({
    ...defaults,
    requestId: '...'
  });
}

<RainbowKitSiweNextAuthProvider getMessageOptions={getMessageOptions}>
```

Or further simplified so that we always merge their options with the defaults.

```tsx
function getMessageOptions() {
  return ({
    requestId: '...'
  });
}

<RainbowKitSiweNextAuthProvider getMessageOptions={getMessageOptions}>
```

The last one might be a step too far, but it's interesting. What do you think?

packages/rainbowkit-siwe-next-auth/src/useSiweNextAuth.ts
markdalgleish: Related to this comment: https://github.com/rainbow-me/rainbowkit/pull/636#discussion_r943067086 — maybe we need to avoid these issues by forcing users to manage these options themselves? Trying to support a minimal setup, but we obviously want to avoid going too far.

markdalgleish: Do we default to `'1'` but support overriding?

packages/example/lib/useConnectionStatus.ts
markdalgleish: I agree, I think it makes sense to avoid giving users too much abstraction in the example. I'll take a look.

packages/rainbowkit-siwe-next-auth/src/useSiweNextAuth.ts
markdalgleish: I've added a new prop called `getSiweMessageOptions` that follows the last example in my previous comment, plus a type called `GetSiweMessageOptions` to make it easier to define this function statically. I've also updated the docs to reflect this. Feeling pretty good about how flexible this is.

packages/rainbowkit-siwe-next-auth/src/useSiweNextAuth.ts
markdalgleish: This is now the default value, but can be overridden if needed via the `getSiweMessageOptions` prop.

markdalgleish:
Re @jxom's earlier comment, I've pushed an update that passes the session object from server to client in the docs and the example apps. We spoke about this offline, but doing this also means that there's no need for the client to cache anything.

nicoglennon (approved):
siweeee

jxom (approved):

PR #639, opened 2022-08-09
fix: disable react-remove-scroll when body overflow is hidden

### LINKED ISSUE #632: [bug] react-remove-scroll-bar: cannot calculate scrollbar size because it is removed

### RainbowKit Version

0.4.5

### wagmi Version

0.6.0

### Current Behavior

We've disabled body scrolling on our page with `overflow: hidden`, since it's a full-page app.

When rainbowkit attempts to hide the scroll bar to show a modal, the react-remove-scroll lib throws an error, which clogs the console (though otherwise does not seem to affect behavior)

`react-remove-scroll-bar: cannot calculate scrollbar size because it is removed (overflow:hidden on body`

image

### Expected Behavior

Should gracefully ignore that body cannot be scrolled, since that's the behavior it would like anyway.

### Steps To Reproduce

add `overflow: hidden` to body

_No response_

### Anything else?

_No response_

markdalgleish:
Fixes #632.

### DIFF
--- a/.changeset/lucky-taxis-poke.md
+++ b/.changeset/lucky-taxis-poke.md
@@ -1,0 +1,5 @@
+---
+'@rainbow-me/rainbowkit': patch
+---
+
+Fix error caused by attempting to prevent page scrolling when the body element's overflow is set to hidden.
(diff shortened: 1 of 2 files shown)


saihaj (approved):

nicoglennon (approved):
🫡

PR #640, opened 2022-08-09
fix: support filtering provider chains

### LINKED ISSUE #631: [bug] Can't remove mainnet from supported chains when configured via `configureChains`

### RainbowKit Version

0.4.5

### wagmi Version

0.6.0

### Current Behavior

Situation: We do not support mainnet in our dApp, but we need to include an RPC-provider in `configureChains` to have ens-name & avatar lookups working properly. So we do something like this:

```ts
export const supportedChains: Chain[] = allChains.filter((chain) =>
  env.supportedChains.includes(chain.id)
)
export const { chains, provider } = configureChains(
  Array.from(new Set([chain.mainnet, ...supportedChains])),
  // ...
```

Now we don't want rainbowkit to offer the mainnet-option in its chain-switcher menu (which it would obviously do if we just use the exported `chains` object). This is why we do filter the `chains` object (again) after running through `configureChains` via:

```ts
export const supportedWagmiChains: Chain[] = chains.filter((chain) => {
  return env.supportedChains.includes(chain.id)
})
```

And using it like:

```tsx
<RainbowKitProvider chains={supportedWagmiChains}>…</RainbowKitProvider>
```

But that doesn't solve the problem. "Ethereum" is still shown as an option in the chain-switch menu (without icon though):

Screen Shot 2022-08-02 at 11 58 28

### Expected Behavior

1. If a chain is not passed to `RainbowKitProvider`, fully ignore it.
2. But more importantly, offer an easy way to provide a mainnet RPC for ens-lookup without requiring mainnet to be supported.

### Steps To Reproduce

_No response_

_No response_

### Anything else?

Thanks in advance as always ❤️

markdalgleish:
Fixes #631.

The main use case for this is to let you configure mainnet for ENS while excluding it from the list of chains available to users when building an L2-only project.

This currently isn't possible because we're using the `chains` prop on `RainbowKitProvider` as the source of truth for whether mainnet has been configured. If you omit mainnet from the chains array passed to RainbowKit then we assume mainnet hasn't been configured in wagmi and bail out on trying to resolve ENS details.

Instead, this PR uses the wagmi provider as the source of truth, getting the list of chains from the provider via the `useProvider` Hook. This allows the list of chains presented in the UI to differ from the chains available to wagmi under the hood.

cc @tmm @jxom I came up with a pretty terse way to manage this without an API change to wagmi, but you still might want to have a look.

### DIFF
--- a/.changeset/lemon-lemons-yawn.md
+++ b/.changeset/lemon-lemons-yawn.md
@@ -1,0 +1,25 @@
+---
+'@rainbow-me/rainbowkit': patch
+---
+
+Support filtering chains before passing them to `RainbowKitProvider`.
+
+This is particularly useful if you're building an L2-only project and you want mainnet to be available for resolving ENS details but you don't want it to be listed in the chain selector.
+
+**Example usage**
+
+This example uses Polygon while supporting ENS from mainnet.
+
+```tsx
+const {
+  chains: [, ...chains], // Omit first chain (mainnet), get the rest
+  provider,
+  webSocketProvider,
+} = configureChains(
+  [chain.mainnet, chain.polygon],
+  [
+    alchemyProvider({ apiKey: '_gg7wSSi0KMBsdKnGVfHDueq6xMB9EkC' }),
+    publicProvider(),
+  ]
+);
+```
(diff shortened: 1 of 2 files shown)


wottpal:
Let me know if I can help get this merged. Really lgtm. 🙂

jxom (approved):
Looks great!!

PR #641, opened 2022-08-10
feat: compact mode

nicoglennon:
compacc mode 👛

https://user-images.githubusercontent.com/16931094/183975907-cc8aa339-31db-4b64-add6-5ba09f92c619.mov


packages/rainbowkit/src/components/ConnectOptions/DesktopOptions.tsx
```diff
@@ -1,28 +1,38 @@
...
-import { ScrollClassName, sidebar } from './DesktopOptions.css';
+import {
+  ScrollClassName,
+  sidebar,
+  sidebarCompactMode,
+} from './DesktopOptions.css';
 
 export enum WalletStep {
   None = 'NONE',
+  ProLearn = 'PRO_LEARN',
```
markdalgleish, line 35: Should be `Compact` now.

Also, tiny nit: I feels like it reads better as `LearnCompact` / `LEARN_COMPACT`.

nicoglennon: agreed. missed this one on the search

packages/rainbowkit/src/components/ConnectOptions/DesktopOptions.tsx
```diff
@@ -174,61 +209,149 @@ export function DesktopOptions({ onClose }: { onClose: () => void }) {
           wallet={selectedWallet}
         />
       );
-      headerLabel = hasQrCode && `Get started with ${selectedWallet.name}`;
+      headerLabel =
```
markdalgleish, line 212: Just confirming, we want to use the short name even when we're not in compact mode?

packages/rainbowkit/src/components/ConnectOptions/DesktopOptions.tsx
```diff
@@ -174,61 +209,149 @@ export function DesktopOptions({ onClose }: { onClose: () => void }) {
...
+        hasQrCode &&
+        `Get started with ${selectedWallet.shortName || selectedWallet.name}`;
```
markdalgleish, line 214: Just confirming, we want to use the short name even when we're not in compact mode?

nicoglennon: yeah - this really is just for coinbase, "coinbase wallet" felt crowded in this screen even on default mode. i can make it the long string on default to keep consistent as much as possible 👍

markdalgleish:
I'm wondering whether we need to avoid using a boolean prop so we don't corner ourselves if we need to add additional modes in the future, e.g. `<RainbowKitProvider modalSize="compact">`.

(I know we already have `coolMode`, but that one's deliberately tongue in cheek 😄)

.changeset/nice-rockets-pay.md
```diff
@@ -0,0 +1,5 @@
+---
+'@rainbow-me/rainbowkit': patch
+---
+
+Add a 'Compact' modal size option for developers looking to use a simpler version of Rainbowkit
```
markdalgleish, line 5:
```suggestion
Add a 'compact' modal size option for developers looking to use a simpler version of RainbowKit, available by setting the `modalSize` prop to `"compact"` on `RainbowKitProvider`.
```

markdalgleish (approved):
Just a small note about adding some more detail to the changeset, but otherwise this looks great 🔥🔥

PR #645, opened 2022-08-11
fix: avoid chain switch on connect if already on supported chain

markdalgleish:
Thanks @jxom for showing me the (currently undocumented) `connector.getChainId()` API which makes this possible 🙏


jxom (approved):
Solid.

markdalgleish:
Confirmed that this doesn't interrupt iOS deep linking too 🙌

kasparkallas:
Thank you, this is much appreciated!

PR #650, opened 2022-08-15
Removed highly opinionated git init, etc on lines 210 to 222

codesport:
1. Lines 210 to 222 cause an install abort on Ubuntu & WSL when using:

```
npm init @rainbow-me/rainbowkit@latest
# or
yarn create @rainbow-me/rainbowkit@latest
# or
pnpm create @rainbow-me/rainbowkit@latest
```
2. It may or may not be good idea to force repo commits on end user's local system. Next.js does it and it works great though!

Therefore, removed highly opinionated `git init` and other actions on lines 210 to 222.  

BTW,  at least for  git on Linux, WSL, and git bash on windows, you'll need to commit to `master` before attempting to rename branch to `main`

Otherwise it will cause an error because there is no reference to HEAD to any branch if nothing as been committed yet.

And, you can try also replicate this by manually creating repo running on Ubuntu or WSL. Try to rename without committing and you'll get a HEAD not found ref error.

Rainbow give this error:

```
  shortMessage: 'Command failed with exit code 128: git branch -m main',
  command: 'git branch -m main',
  escapedCommand: 'git branch -m main',
  exitCode: 128,
  signal: undefined,
  signalDescription: undefined,
  stdout: '',
```

Which is because of the head issue and renaming a branch with 0 commits


markdalgleish:
Rather than omitting the git init/commit step, I've opted to remove the branch rename command in https://github.com/rainbow-me/rainbowkit/pull/765.

markdalgleish:
We no longer rename the branch in [create-rainbowkit v0.1.2.](https://github.com/rainbow-me/rainbowkit/releases/tag/%40rainbow-me%2Fcreate-rainbowkit%400.1.2)

PR #652, opened 2022-08-16
Allow internal column scrolling on shorter screens

kolber:
Minor css rework to allow the connect modal internals to scroll when the screen is shorter than the fixed modal height.

Before (clipping off-screen):
Screen Shot 2022-08-16 at 2 36 28 pm

After (internal column scrolling):
Screen Shot 2022-08-16 at 2 41 35 pm

https://user-images.githubusercontent.com/10325/184798611-9b42b40e-e2cf-41e5-acc4-b694e95dcd47.mov

### DIFF
--- a/packages/rainbowkit/src/components/ConnectOptions/DesktopOptions.tsx
+++ b/packages/rainbowkit/src/components/ConnectOptions/DesktopOptions.tsx
@@ -182,7 +182,11 @@ export function DesktopOptions({ onClose }: { onClose: () => void }) {
   }
 
   return (
-    <Box display="flex" flexDirection="row" style={{ maxHeight: 504 }}>
+    <Box
+      display="flex"
+      flexDirection="row"
+      style={{ maxHeight: '100vh', overflowY: 'auto' }}
+    >
       <Box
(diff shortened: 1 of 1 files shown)


markdalgleish:
Thanks for the PR! I just had a look at the preview deployment at it looks like there's a few layout issues introduced.

On the injected connector screen, the content has shifted to the top:

Screen Shot 2022-08-17 at 11 15 38 am

Same on the "Get a wallet" screen:

Screen Shot 2022-08-17 at 11 20 29 am

On the QR code screen. the content too wide and causes the close icon to get clipped and the modal to scroll horizontally:

Screen Shot 2022-08-17 at 11 15 45 am

Also on the intro screen, the content is higher than it was previously.

Before:

Screen Shot 2022-08-17 at 11 19 43 am

After:

Screen Shot 2022-08-17 at 11 15 58 am

kolber:
Thanks @markdalgleish! It looks like the compact mode fixes the broken responsive styles, so we've switched to that on our end. :)

Closing this PR as unnecessary. 👍

PR #662, opened 2022-08-18
chore: change rainbowkit-siwe-next-auth version to 0.1.0

markdalgleish:
Just noticed Changesets wanted to publish as v1 even though it was marked as minor. This PR manually drops the version, after which it should publish as v0.1.0.


nicoglennon (approved):
🛹

PR #675, opened 2022-08-22
feat: expose useRecentTransactions hook

spennyp:


spennyp:
Adds feature discussed here: https://github.com/rainbow-me/rainbowkit/discussions/373

shrugs:
++ access to the rainbowkit transaction store would be a nice upgrade for my project — the tx store does exactly what i want, (although would also like to be able to annotate txs with additional info so i can filter them app-side) so i've resorted to copying the TransactionStore itself into my project and writing my own hooks, then dual-publishing to rainbow's internal store.

smakosh:
> ++ access to the rainbowkit transaction store would be a nice upgrade for my project — the tx store does exactly what i want, (although would also like to be able to annotate txs with additional info so i can filter them app-side) so i've resorted to copying the TransactionStore itself into my project and writing my own hooks, then dual-publishing to rainbow's internal store.

About to do the same thing. Feels bad having the same code around when the solution needs just to be exported.

smakosh (approved):

aj-may:
Can/should we add docs for this? 📚

DanielSinclair:
Thanks for the contribution @spennyp

PR #688, opened 2022-08-26
feat: download options for mobile & extension wallets

nicoglennon:
Screen Shot 2022-08-29 at 1 26 13 PM

adds ability to download both extension and mobile version of wallets who have both


peduarte:
yooo love this!

noticed a couple of things on Safari:
- the gradient isnt rendering properly
- it's a bit weird that clicking the "add to browser" button takes me to the google chrome store
- using the term "Browser" is also weird imo. in Chrome we use "Chrome", same in Brave. so im guessing this is because there is no such extension for Safari? if so, should we consider not rendering the card at all? cc @mikearndt 

CleanShot 2022-08-30 at 19 49 44@2x

nicoglennon:
thanks @peduarte and welcome back 🫡🫡🫡

> * the gradient isnt rendering properly
is this the colors being a little faded? i just noticed that. i had rearranged the filtering and seems like it looks slightly different in safari
> * using the term "Browser" is also weird imo. in Chrome we use "Chrome", same in Brave. so im guessing this is because there is no such extension for Safari? if so, should we consider not rendering the card at all?
yes, we don't use safari because the extension doesn't exist in safari, so we just say "browser". it's more a fallback for anything else (opera, etc), but i see what you mean about directing to the chrome store. one positive thing about this method is ppl could see that they need chrome to use these extensions, which the chrome store lets you know. i don't think removing the browser option is better than showing how this exists on another browser

peduarte:
@nicoglennon nice one, happy to go with the generic "browsers" term then 🚀 

about the gradient, first time i opened it, there was no gradient just a harsh line. but after taking the screenshot the gradient came out smooth again... so not actually sure what happened there. maybe it was a rendering bug related to hardware acc. or something 🤔

packages/rainbowkit/src/components/ConnectOptions/ConnectDetails.tsx
```diff
@@ -322,6 +348,186 @@ export function ConnectDetail({
...
+        flexDirection="row"
+        gap="24"
+        height="max"
+        justifyContent="center"
+        style={{ zIndex: 1 }}
+      >
+        <Box>
+          <AsyncImage
+            height="60"
+            src={iconUrl}
```
nicoglennon, line 430: browser svg logo fed into async image

packages/rainbowkit/src/components/ConnectOptions/ConnectDetails.tsx
```diff
@@ -1,26 +1,46 @@
...
 import { CreateIcon, preloadCreateIcon } from '../Icons/Create';
+import { preloadRefreshIcon, RefreshIcon } from '../Icons/Refresh';
 import { preloadScanIcon, ScanIcon } from '../Icons/Scan';
-import { SpinnerIcon } from '../Icons/Spinner';
 import { QRCode } from '../QRCode/QRCode';
+import { ModalSizeContext } from '../RainbowKitProvider/ModalSizeContext';
 import { Text } from '../Text/Text';
 import { WalletStep } from './DesktopOptions';
 
+const getBrowserSrc: () => Promise<string> = async () => {
```
nicoglennon, line 22: getBrowserStc function to only load the correct browser icon

nicoglennon:
> @nicoglennon nice one, happy to go with the generic "browsers" term then 🚀
> 
> about the gradient, first time i opened it, there was no gradient just a harsh line. but after taking the screenshot the gradient came out smooth again... so not actually sure what happened there. maybe it was a rendering bug related to hardware acc. or something 🤔

fixed the gradient issue in the latest. was a safari performance issue 🐢

peduarte:
heck yeah, looking crisp! 💯

peduarte (approved):

PR #726, opened 2022-09-09
Resolves #722 Add example with astro

### LINKED ISSUE #722: [bug] QRCode not showing up

### RainbowKit Version

0.5.1

### wagmi Version

0.6.4

### Current Behavior

Popup works, Metamask connects without problem, but no QR code for wallet connect and wallet link.
Screen Shot 2022-09-07 at 10 06 12 PM
Screen Shot 2022-09-07 at 10 06 27 PM

### Expected Behavior

QR code showing up when selected

### Steps To Reproduce

Used with a react wrapper in astro, with both client:load and client:only="react"

https://lensmo-fe.vercel.app/

### Anything else?

_No response_

mikelxc:
#722

### DIFF
--- a/examples/with-astro/.gitignore
+++ b/examples/with-astro/.gitignore
@@ -1,0 +1,20 @@
+# build output
+dist/
+.output/
+
+# dependencies
+node_modules/
+
+# logs
+npm-debug.log*
+yarn-debug.log*
+yarn-error.log*
+pnpm-debug.log*
+
+
+# environment variables
+.env
+.env.production
+
+# macOS-specific files
+.DS_Store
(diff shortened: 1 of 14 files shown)


markdalgleish:
I tried running this and couldn't see a connect button. There was an error in the console — seemed related to the polyfills not being applied?

mikelxc:
image

The kit compiles and runs without a problem on my end. 
You will need to create a yarn.lock file and do an yarn install

nickbytes:
@mikelxc thanks for taking the time to check out RainbowKit! I'd prefer we limit the number of example applications in the repo that we have to maintain, but feel free to publish to your own repo! If astro absolutely explodes in popularity over the next year and we see many people having issues getting setup, we can talk about adding this.

Issue #461, opened 2022-06-02
[bug] Package throws import errors when running jest test suites


brightiron:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### RainbowKit Version

0.2.2

### wagmi Version

0.4.6

### Current Behavior

Package causes jest to throw a 'Cannot use import statement outside a module' error 

Hello Sers! 
Digging this project and appreciate all you guys are doing. 
Running into an issue when running our test suite where this package throws import errors. There are some things we can do on our side to workaround (i.e. mock import, transformIgnorePatterns, etc) but this is not something we've had to do for any other dependency.

To rule out our project specifically i've created a fresh create-react-app, installed rainbowkit, wagmi, and ethers. When running `yarn test` we experience the exact same behavior. 

I imagine this is something others will soon encounter. 

```
    Jest encountered an unexpected token

    This usually means that you are trying to import a file which Jest cannot parse, e.g. it's not plain JavaScript.

    By default, if Jest sees a Babel config, it will use that to transform your files, ignoring "node_modules".

    Here's what you can do:
     • If you are trying to use ECMAScript Modules, see https://jestjs.io/docs/en/ecmascript-modules for how to enable it.
     • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
     • If you need a custom transformation specify a "transform" option in your config.
     • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.

    You'll find more details and examples of these config options in the docs:
    https://jestjs.io/docs/en/configuration.html

    Details:

    /rainbowkit-jest-test/node_modules/@rainbow-me/rainbowkit/dist/index.js:1
    ({"Object.<anonymous>":function(module,exports,require,__dirname,__filename,global,jest){import {
```

### Expected Behavior

Package should not throw import errors when running jest test suite and shouldn't have to mock package import.  

### Steps To Reproduce

Checkout: https://github.com/brightiron/rainbowkit-jest-test (Clean create-react-app project with only wagmi, rainbowkit and ethers installed)

Run:
```yarn install```
``` yarn test```

### Link to Minimal Reproducible Example (CodeSandbox, StackBlitz, etc.)

https://github.com/brightiron/rainbowkit-jest-test

nickbytes:
Thanks for the detailed issue and minimal reproducible example @brightiron – we'll check it out!

brightiron:
[...]

PR #727, opened 2022-09-09
fix: add `styles.css/package.json` as an alias for Jest

markdalgleish:
Some older build/test tools have issues when importing RainbowKit styles due to lack of support for the `exports` field, which we previously used to publicly alias `@rainbow-me/rainbowkit/dist/index.css` as `@rainbow-me/rainbowkit/styles.css`.

To fix this, we now copy the CSS file into the root of the package as part of our build process so that it can be imported without an alias.

Note that I've left the original `dist/index.css` file there since some consumers were importing this file directly and I don't want to break them.

Related issue: https://github.com/rainbow-me/rainbowkit/issues/461

Fixes: PAR-18

### DIFF
--- a/.changeset/beige-clocks-cross.md
+++ b/.changeset/beige-clocks-cross.md
@@ -2,8 +2,6 @@
 '@rainbow-me/rainbowkit': patch
 ---
 
-Fix the `@rainbow-me/rainbowkit/styles.css` import for tooling that doesn't support the `exports` field in `package.json`.
+Fix the `@rainbow-me/rainbowkit/styles.css` import for Jest, and other tooling that doesn't support the `exports` field in `package.json`.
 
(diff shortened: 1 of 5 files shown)


peduarte (dismissed):
interesting stuff! king of builds

markdalgleish:
I just pushed an update based on a genius idea by @jxom that simplifies things a lot.

We no longer need to copy files around—instead, there's a `styles.css` _directory_ that contains a `package.json` file with a `main` field pointed at `"../dist/index.css"`. This way Jest resolves the `/styles.css` import to this `package.json` file, and is then redirected to the generated CSS output. I've tested this in Jest and it works. Really clever.

jxom (approved):
Amazing!

peduarte:
daaamnnnn, i love that!

PR #737, opened 2022-09-10
docs: ensure selection color always updates to new value

frantic:


peduarte (commented):
hey, thanks for this PR!

personally getting the same color doesnt bother me. i actually quite like it. but i'm also fine with this change

i'm a bit concerned with the usage of generators though. its not a pattern we use. i'd prefer if we created some type of "shuffle array" function in `/lib/shuffleArray` and then used that. preferably outside of the effect

could be based on the [Fisher Yates shuffle](http://en.wikipedia.org/wiki/Fisher-Yates_shuffle) , that ensures no item is ever in the same index
```ts
// lib/shuffleArray.ts
const shuffleArray = (array: string[]) => {
  for (let i = array.length - 1; i > 0; i--) {
    const j = Math.floor(Math.random() * i);
    const temp = array[i];
    array[i] = array[j];
    array[j] = temp;
  }
  return array;
};
```

then replace `getColor` with:
```ts
// _app.tsx
const getColor = () => shuffleArray(highlightColors)[0];
```

peduarte (approved):

peduarte:
mikey really wants this, so lets get it in. we can refactor later if we want to stay away from generators. thanks ❄️

PR #738, opened 2022-09-12
feat: keep wallets in their group, ordered by most recently used

peduarte:
Resolves PAR-13

---

~This PR introduces a `disableRecentWallets` prop to completely disable the re-organisation of recently used wallets on desktop and mobile modals.~

~Discussion is still ongoing to see if there's potentially a better solution. for this reason this PR is in Draft mode.~

~Follow the discussion here: https://github.com/rainbow-me/rainbowkit/discussions/625#discussioncomment-3625668~

---

UPDATE:

This PR implements "solution C" as [discussed here](https://github.com/rainbow-me/rainbowkit/discussions/625#discussioncomment-3625668).

> solution C: disable "recent group" on desktop, but re-order the wallet list
basically means desktop having the same behaviour as mobile, where most recently used wallets are first in the list

**Before**

image

**After**

CleanShot 2022-09-15 at 14 52 49@2x


packages/rainbowkit/src/components/ModalSelection/ModalSelection.tsx
```diff
@@ -85,7 +88,19 @@ export const ModalSelection = ({
...
             />
-            <div>{name}</div>
+            <Box>
+              <Box style={{ marginTop: recent ? -2 : undefined }}>{name}</Box>
+              {recent && (
+                <Text
+                  color="accentColor"
+                  size="12"
+                  style={{ lineHeight: 1, marginTop: -1 }}
+                  weight="medium"
```
peduarte, line 98: couple of inline styles here and line 92. this is because to ensure consistent vertical gaps between all items, the computed height of the button needs to be 28 even if it includes the "recent" label.

however, the label was causing the computed height to be 32 (3px too high).

setting negative margin only in the "recent" label, meant that both texts were too tight together. for this reason, we're offsetting the wallet label by 2, and the recent label by 1. 

image

peduarte: approved by @mikearndt

markdalgleish:
I noticed the accent color text is hidden when the wallet is selected since we use the accent color as the background:

Screen Shot 2022-09-19 at 7 04 25 am

We'll probably need to set it to `accentColorForeground` when selected?

markdalgleish:
Is it expected that the entire group of the most recently used wallet moves to the top of the list? My concern with this is that it could push other groups off the screen (especially as we add more wallets), and those groups could also contain other recently used wallets.

e.g. in this screenshot, "Other" has taken precedence over "Popular":

Screen Shot 2022-09-19 at 7 10 01 am

jxom:
Yeah, I agree with @markdalgleish on the ordering of wallet groups. Apart from that, I am pretty happy with this PR!

peduarte:
@markdalgleish good catch, i missed that! 

CleanShot 2022-09-19 at 09 48 00@2x


about the group ordering, the idea there was to really push the most recently used wallet to the top, and that includes the group.

but i can see how re-ordering the group is kinda weird. let me make some changes to only re-order wallets, but keep groups in place.

jxom (approved):
Tested again and looking good!

markdalgleish (approved):

markdalgleish:
Nice one! 🎉

PR #741, opened 2022-09-12
fix: update download card designs

nicoglennon:
gradient changes


packages/rainbowkit/src/css/sprinkles.css.ts
```diff
@@ -25,6 +25,8 @@ const themeContractValues = {
     connectButtonText: '',
     connectButtonTextError: '',
     connectionIndicator: '',
+    downloadBottomCardBackground: '',
+    downloadTopCardBackground: '',
```
markdalgleish, line 29: These new properties will need to go in the custom theme docs.

packages/rainbowkit/src/utils/colors.ts
```diff
@@ -0,0 +1,30 @@
+export const convertHexToRGBA = (
+  hexCode: string,
```
markdalgleish, line 2: If we need colors to be provided in a hex format, we might need to add some validation when wallets are initially passed to RainbowKit, otherwise this code will silently break.

packages/rainbowkit/src/wallets/Wallet.ts
```diff
@@ -33,6 +33,7 @@ type RainbowKitConnector<C extends Connector = Connector> = {
 };
 
 export type Wallet<C extends Connector = Connector> = {
+  accentColor?: string;
```
markdalgleish, line 36: I wonder if this should be called `iconAccent` so that it forms a pair with `iconBackground`? This might also avoid potential confusion that it's used as a general accent color in the UI.

packages/rainbowkit/src/components/ConnectOptions/ConnectDetails.tsx
```diff
@@ -373,10 +376,13 @@ const DownloadOptionsBox = ({
   iconUrl: string | (() => Promise<string>);
   iconBackground?: string;
 }) => {
+  const isBrowserCard = !accentColor;
```
markdalgleish, line 379: Since the `accentColor` property is optional on the `Wallet` type, this means some wallets could inadvertently set `isBrowserCard` to true? Would it maybe be safer to have an explicit prop for this rather than inferring it? e.g. `type="browser"` / `type="app"`?

markdalgleish:
Just noticed, this will also need a changeset.

packages/rainbowkit/src/wallets/Wallet.ts
nicoglennon: gotcha, will rename for clarity

packages/rainbowkit/src/utils/colors.ts
nicoglennon: i added some validation in `connectorsForWallets.ts`, will tag u there

.changeset/gorgeous-humans-fold.md
```diff
@@ -0,0 +1,5 @@
+---
+'@rainbow-me/rainbowkit': patch
+---
+
+The new Wallet installation flows add two new colors to RainbowKit's `colors` property in the custom `theme` object: `downloadBottomCardBackground` and `downloadTopCardBackground`. Additionally, we now include the optional `iconAccent` prop in the `Wallet` object, specifically for wallets with both browser extensions, to use as the accent color for the new download installation flows.
```
markdalgleish, line 5: Was this sentence incomplete?

```suggestion
The new Wallet installation flows add two new colors to RainbowKit's `colors` property in the custom `theme` object: `downloadBottomCardBackground` and `downloadTopCardBackground`. Additionally, we now include the optional `iconAccent` prop in the `Wallet` object, specifically for wallets with both a browser extension and an app, to use as the accent color for the new download installation flows.
```

markdalgleish (approved):

PR #744, opened 2022-09-13
chore(changesets): only update peers when out of range

markdalgleish:
I've noticed that we're inadvertently publishing a new version of `rainbowkit-siwe-next-auth` whenever `rainbowkit` changes, even though it still satisfies the same version range.

Hopefully this super secret Changesets config option does the trick, borrowed from [vanilla-extract's Changesets config](https://github.com/seek-oss/vanilla-extract/blob/master/.changeset/config.json) 🙏


jxom (approved):
yesss

PR #747, opened 2022-09-13
chore: add dark theme page backgrounds to example app

nicoglennon:
Screen Shot 2022-09-13 at 2 30 22 PM

### DIFF
--- a/packages/example/pages/index.tsx
+++ b/packages/example/pages/index.tsx
@@ -137,6 +137,8 @@ const Example = ({ authEnabled }: AppContextProps) => {
         style={{
           display: 'flex',
           justifyContent: 'flex-end',
+          position: 'sticky',
+          top: 8,
         }}
       >
         <ConnectButton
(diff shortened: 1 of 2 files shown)


markdalgleish:
Nice one! 🎉

I noticed that if the window is taller than the page content, you can see the white body background underneath. Could you fix this?

Screen Shot 2022-09-14 at 8 36 07 am

packages/example/pages/index.tsx
```diff
@@ -137,6 +137,8 @@ const Example = ({ authEnabled }: AppContextProps) => {
         style={{
           display: 'flex',
           justifyContent: 'flex-end',
+          position: 'sticky',
+          top: 8,
```
peduarte, line 141: legend

nicoglennon:
> I noticed that if the window is taller than the page content, you can see the white body background underneath. Could you fix this?

yup missed this when working out of my laptop, should be fixed now

markdalgleish (approved):

PR #753, opened 2022-09-14
feat: add prop to specify ens chain

danielr18:
Allow changing which chain is used for ENS, by setting the `ensChain` prop on `RainbowKitProvider`

Relates to: https://github.com/rainbow-me/rainbowkit/discussions/752


.changeset/rude-fireants-doubt.md
```diff
@@ -0,0 +1,6 @@
+---
+'example': patch
```
markdalgleish, line 2: This should be removed.

packages/example/pages/_app.tsx
```diff
@@ -380,6 +382,30 @@ function RainbowKitApp({ Component, pageProps }: AppProps) {
                         </select>
                       </td>
                     </tr>
+                    <tr>
+                      <td>ensChain</td>
```
markdalgleish, line 386: Thanks for adding this!

markdalgleish:
Thanks for the PR!

This will need some docs on the "Chains" page: https://www.rainbowkit.com/docs/chains — did you want to write these, or would you prefer push updates to your PR?

.changeset/rude-fireants-doubt.md
danielr18: Removed

danielr18:
Thanks, I've added a section in the docs.

DanielSinclair (commented):
A bit unsure of the API design here. It may be a better UX for RainbowKit to resolve ENS for the user's selected chain or `initialChain`, as opposed to bubbling up an ENS-specific prop to the dApp dev. Can you walk me through the design decisions?

DanielSinclair:
Will close this for now because there might be a better/simpler means to an end for this particular testnet ENS problem. Can reopen if this changes or based on additional conversations here.

PR #754, opened 2022-09-14
feat: add Infinity Wallet

infinitys2025:


packages/rainbowkit/src/wallets/walletConnectors/infinityWallet/infinityWallet.ts
```diff
@@ -0,0 +1,98 @@
+/* eslint-disable sort-keys-fix/sort-keys-fix */
+import { openInfinityWallet } from '@infinitywallet/infinity-connector';
```
markdalgleish, line 2: Since this is using wagmi's injected connector, can we avoid adding another dependency?

markdalgleish:
We've just re-worked RainbowKit internals in v0.7 so that wallets can be imported individually, ensuring our bundle size doesn't keep increasing as we continue to add more wallets.

Are you able to give us write access to your fork so that we can get you updated to the new API and apply some changes before merging?

markdalgleish:
Sorry for the delay on this, and thanks for the PR!

We've just re-worked RainbowKit internals in v0.7 so that wallets can be imported individually, ensuring our bundle size doesn't keep increasing as we continue to add more wallets.

Thanks for giving us access to your fork — I've pushed some updates to your PR migrating it to the new API.

packages/rainbowkit/src/wallets/walletConnectors/infinityWallet/infinityWallet.ts
InfinityWallet: Hi @markdalgleish, as Infinity Wallet is not a browser extension but the first standalone web3 browser wallet, the function `openInfinityWallet` is used both to detect if Infinity Wallet is installed on click and to open the DApp in the Infinity Wallet from any browser.

If you would like, rather than importing the dependency for this we could directly integrate the code into the repository?

Is there anything else required to merge this?

InfinityWallet:
> Sorry for the delay on this, and thanks for the PR!
> 
> We've just re-worked RainbowKit internals in v0.7 so that wallets can be imported individually, ensuring our bundle size doesn't keep increasing as we continue to add more wallets.
> 
> Thanks for giving us access to your fork — I've pushed some updates to your PR migrating it to the new API.

Hi @markdalgleish, we we wondering if there was anything else needed to merge this PR? Thanks

packages/rainbowkit/package.json
```diff
@@ -60,6 +60,7 @@
     "vitest": "^0.5.0"
   },
   "dependencies": {
+    "@infinitywallet/infinity-connector": "^1.0.5",
```
DanielSinclair, line 63: Our current approach here is to rely solely on `@wagmi/connectors` for peer dependency connectors. Please first work with the `wagmi` team for upstream support, and then we'll revisit. https://github.com/wagmi-dev/references/tree/main/packages/connectors

DanielSinclair (changes requested):

packages/rainbowkit/src/wallets/walletConnectors/infinityWallet/infinityWallet.ts
```diff
@@ -0,0 +1,89 @@
...
+          }),
+        };
+      }
+
+      const connector = getWalletConnectConnector({ chains });
+
+      return {
+        connector,
+        mobile: {
+          getUri: undefined,
```
DanielSinclair, line 53: Please also provide clarity on this `getUri`, and the desktop equivalent. These should not return undefined for WalletConnect-supporting wallets.

DanielSinclair:
Closing this for now. Can revisit if there is a better integration path.

PR #756, opened 2022-09-15
feat: automatically hide Injected Wallet option

markdalgleish:
To make it easier for consumers to include the Injected Wallet fallback, this PR internalises the logic for calculating whether the option should be visible or not. This means that consumers no longer need to do this:

```ts
const needsInjectedWalletFallback =
  typeof window !== 'undefined' &&
  window.ethereum &&
  !window.ethereum.isMetaMask &&
  !window.ethereum.isCoinbaseWallet;
  
const connectors = connectorsForWallets([
  {
    groupName: 'Suggested',
    wallets: [
      ...(needsInjectedWalletFallback ? [wallet.injected({ chains })] : []),
      ...etc,
    ],
  },
]);
```

Instead, they can now simply add `wallet.injected({ chains })` to the list like any other wallet. The old code will still work, it's just redundant, so this isn't a breaking change.

I've also updated the custom wallet documentation to prioritise `wallet.injected` so that it (hopefully) doesn't get forgotten when consumers set up custom wallet lists.

Note that this PR also fixes PAR-15 by moving the Injected Wallet option to the start of the list since it's only visible when necessary.

This feature required a few changes that each have their own changeset, so you should find them useful for getting an idea of the moving parts.

### DIFF
--- a/packages/rainbowkit/src/wallets/getDefaultWallets.ts
+++ b/packages/rainbowkit/src/wallets/getDefaultWallets.ts
@@ -4,7 +4,7 @@ import { connectorsForWallets } from './connectorsForWallets';
 import { brave } from './walletConnectors/brave/brave';
 import { coinbase } from './walletConnectors/coinbase/coinbase';
 import { injected } from './walletConnectors/injected/injected';
-import { isMetaMask, metaMask } from './walletConnectors/metaMask/metaMask';
+import { metaMask } from './walletConnectors/metaMask/metaMask';
 import { rainbow } from './walletConnectors/rainbow/rainbow';
 import { walletConnect } from './walletConnectors/walletConnect/walletConnect';
 
@@ -18,25 +18,16 @@ export const getDefaultWallets = ({
   connectors: ReturnType<typeof connectorsForWallets>;
   wallets: WalletList;
 } => {
-  const needsInjectedWalletFallback =
-    typeof window !== 'undefined' &&
-    window.ethereum &&
-    !isMetaMask(window.ethereum) &&
-    !window.ethereum.isCoinbaseWallet &&
-    !window.ethereum.isBraveWallet;
-
   const wallets: WalletList = [
     {
       groupName: 'Popular',
       wallets: [
+        injected({ chains, shimDisconnect: true }),
         rainbow({ chains }),
         coinbase({ appName, chains }),
         metaMask({ chains, shimDisconnect: true }),
         walletConnect({ chains }),
         brave({ chains, shimDisconnect: true }),
-        ...(needsInjectedWalletFallback
-          ? [injected({ chains, shimDisconnect: true })]
-          : []),
       ],
     },
   ];
--- a/packages/rainbowkit/src/wallets/getDefaultWallets.ts
+++ b/packages/rainbowkit/src/wallets/getDefaultWallets.ts
@@ -22,12 +22,12 @@ export const getDefaultWallets = ({
     {
       groupName: 'Popular',
       wallets: [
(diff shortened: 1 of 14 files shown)


packages/rainbowkit/src/wallets/getDefaultWallets.ts
```diff
@@ -18,25 +18,16 @@ export const getDefaultWallets = ({
   connectors: ReturnType<typeof connectorsForWallets>;
   wallets: WalletList;
 } => {
-  const needsInjectedWalletFallback =
```
jxom, line 21: neat

packages/rainbowkit/src/wallets/walletConnectors/injected/injected.ts
```diff
@@ -16,6 +16,13 @@ export const injected = ({
   name: 'Injected Wallet',
   iconUrl: async () => (await import('./injected.png')).default,
   iconBackground: '#fff',
+  hidden: ({ wallets }) =>
```
jxom, line 19: so clean

site/data/docs/custom-wallet-list.mdx
```diff
@@ -60,78 +61,75 @@ const App = () => (
 
 ### Built-in wallets
 
-The following wallets are provided via the `wallet` object (in alphabetical order).
+The following wallets are provided via the `wallet` object.
 
-#### Argent
+#### Generic Wallets
```
jxom, line 66: Yeah, I really like this separation between wallets: "Generic" vs. "Specific". Nice.

packages/rainbowkit/src/wallets/walletConnectors/metaMask/metaMask.ts
```diff
@@ -10,7 +10,7 @@ export interface MetaMaskOptions {
   shimDisconnect?: boolean;
 }
 
-export function isMetaMask(ethereum: NonNullable<typeof window['ethereum']>) {
+function isMetaMask(ethereum: NonNullable<typeof window['ethereum']>) {
```
jxom, line 13: FYI, this is out of sync with wagmi atm... We may need to either add [tocken pocket](https://github.com/wagmi-dev/wagmi/blob/main/packages/core/src/connectors/metaMask.ts#L122) or find a way to keep in sync with wagmi.

site/data/docs/custom-wallet-list.mdx
```diff
@@ -267,8 +256,9 @@ const connectors = connectorsForWallets([
   {
     groupName: 'Others',
     wallets: [
-      wallet.walletConnect({ chains }),
+      wallet.injected({ chains, shimDisconnect: true }),
```
jxom, line 259: `shimDisconnect` is already default truthy.

```suggestion
      wallet.injected({ chains }),
```

jxom: Maybe we can omit other usages while we are at it?

jxom (approved):
🚀

site/data/docs/custom-wallet-list.mdx
markdalgleish: Ah, nice catch. Done 👍

packages/rainbowkit/src/wallets/walletConnectors/metaMask/metaMask.ts
markdalgleish: Syncing up with wagmi is a good idea, but in the meantime I've updated our logic in this PR.

jxom (approved):
ship it

peduarte:
love it

PR #769, opened 2022-09-20
docs: fix Remix example, clean up old react-query dependency

markdalgleish:
While fixing the NFT mint example, I noticed that the Remix example was missing polyfills and had some outdated code/patterns, so I've applied the following updates:

- Added the `buffer` package as a polyfill, aliased as `buffer-polyfill` because Remix complains if you try to use a package with the same name as a Node built-in. This fixes the WalletConnect integration. Note that I had to mark the polyfills file as a side effect in package.json, otherwise it got trimmed from the build.
- Renamed the `alchemyId` option to `apiKey`.
- Moved the wagmi client setup code into a lazy `useState` call rather than a `useEffect` so we don't need to manually annotate the types. This is a nice clean up because we were depending on an older version of react-query, which I've removed in this PR.

Note that there's still a bug with the Coinbase Wallet SDK, but that seems to be an issue with their library being used in Remix due to one of their dependencies importing from `events/` rather than `events`, so it's not something we have control over. I can follow up with them separately to see if we can get that fixed.


examples/with-remix/app/root.tsx
```diff
@@ -53,12 +60,41 @@ export default function App() {
...
   // Remix modules cannot have side effects so the initialization of `wagmi`
-  // client is abstracted away in a `useEffect`.
+  // client happens during render, but the result is cached via `useState`
+  // and a lazy initialization function.
   // See: https://remix.run/docs/en/v1/guides/constraints#no-module-side-effects
-  const { client, chains } = useSetupWagmi({
-    appName: 'RainbowKit Remix Example',
-    alchemyId: ENV.ALCHEMY_ID,
-    enablePublicTestnets: ENV.PUBLIC_ENABLE_TESTNETS,
+  const [{ client, chains }] = useState(() => {
```
jxom, line 66: Is there a reason to use `useState` over `useMemo`?

markdalgleish: The initializer for `useState` is guaranteed to only run once, while `useMemo` doesn't technically guarantee this.

From https://reactjs.org/docs/hooks-reference.html#usememo:
> **You may rely on useMemo as a performance optimization, not as a semantic guarantee.** In the future, React may choose to “forget” some previously memoized values and recalculate them on next render, e.g. to free memory for offscreen components. Write your code so that it still works without `useMemo` — and then add it to optimize performance.

jxom: True!

jxom (approved):

PR #771, opened 2022-09-20
fix: include `styles.css` in npm bundle

rzadp:
Follow up to https://github.com/rainbow-me/rainbowkit/pull/727

@markdalgleish The change was great but the folder is not included in the NPM bundle - it works in the monorepo but does not work for outside projects.

Screenshot 2022-09-20 at 11 21 34

Thanks for your commitment to support different build tools.

### DIFF
--- a/.changeset/large-rabbits-provide.md
+++ b/.changeset/large-rabbits-provide.md
@@ -1,0 +1,5 @@
+---
+'@rainbow-me/rainbowkit': patch
+---
+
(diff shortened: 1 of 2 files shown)


markdalgleish (approved):
Thank you!

markdalgleish:
Thanks again! Fix is out now in [v0.6.1](https://github.com/rainbow-me/rainbowkit/releases/tag/%40rainbow-me%2Frainbowkit%400.6.1).

PR #783, opened 2022-09-28
feat: docs search

peduarte:
Docs with Algolia. A bit of a custom implementation so we can customise the search button and the modal. Overrides are as global CSS, and due to the lack of control over the markup, it's not really the cleanest approach. But visually looks good! 

UI/design approved by Mikey.

CleanShot 2022-09-29 at 10 01 12@2x

Note: The favourite/remove functionality isnt working. I reached out to algolia but no response yet. I believe it will work only on the production url.


site/components/Search/Search.tsx
```diff
@@ -0,0 +1,178 @@
...
+  useCallback,
+  useContext,
+  useEffect,
+  useState,
+} from 'react';
+import { createPortal } from 'react-dom';
+
+const INDEX_NAME = 'rainbowkit';
+const API_KEY = '671f4a1482dd79c7a57883c939250293';
+const APP_ID = 'JY91C5WNW9';
```
peduarte, line 19: these are all public keys, so fine to be here

site/components/Button/Button.tsx
```diff
@@ -32,7 +32,7 @@ type BaseProps = {
   >;
 
 type WithAnchor = {
-  as?: 'a';
+  as?: 'a' | 'span';
```
markdalgleish, line 35: Should this be on the `WithoutAnchor` type? Otherwise the types say it's valid to create a span with an `href` attribute.

markdalgleish:
I noticed that the way we're handling section names within each page is causing search results like this:

Screen Shot 2022-10-04 at 11 21 17 am

Screen Shot 2022-10-04 at 11 21 25 am

site/components/Button/Button.tsx
peduarte: makes sense!

peduarte:
@markdalgleish the quality of the results is something we need to configure within the crawler. the crawler only crawls the production site. this PR has made a few small changes to offer extra selectors for the crawler, so it should get better

markdalgleish (approved):

PR #801, opened 2022-10-07
feat: show erc-20 token balance of the user's choosing

owieth:
Implementation for https://github.com/rainbow-me/rainbowkit/discussions/409


nickbytes:
@owieth this could be achieved by implementing a [Custom ConnectButton](https://www.rainbowkit.com/docs/custom-connect-button), is there any reason we shouldn't push people to do this vs adding more configuration here?

owieth:
@nickbytes Can't see how a Custom ConnectButton Implementation could change the displayed token in the account modal:
image

nickbytes:
@owieth my bad, I misunderstood what you were saying. I thought you meant you wanted to replace the balance in the connected state.

So if you wanted to replace this in the modal, you _could_ actually use a Custom ConnectButton, but there's a little more work to be done. In using the [Custom ConnectButton example](https://www.rainbowkit.com/docs/custom-connect-button), you can see the following:

image

If you replaced `openAccountModal` modal with your own interaction, you could display a modal completely of your choosing and ditch RainbowKit's altogether, giving you ultimate flexibility. We've seen people (like backed.xyz) implement this style to pretty cool effect. You can do things like instead navigate to an Account Page (like Backed does vs. using modals) or create your own modal that may show more/less information that is specific to your application.

For the time being, I'd suggest this option. I'll discuss w/ the team a bit and see how common of a use case this might be and whether we should implement.

owieth:
Hey @nickbytes 
Sorry for replying late...
Ahhh didn't know that you could do that, is this referenced somewhere in the docs? 
Definitely going to try out your approach

(let's have this pr open until the team comes to a decision?)

ChrisClayton:
> If you replaced `openAccountModal` modal with your own interaction, you could display a modal completely of your choosing and ditch RainbowKit's altogether.

Rainbowkit is a good project that I wouldn't want to ditch. With this pull-request implemented it'd go from "good" to "great".

Correct me if I'm wrong but If my understanding is correct, the core team specifically chose to allow Rainbowkit to be extensible. Rainbowkit currently implements all the wallets, and all the chains and all the providers. Adding new ones doesn't require ditching any builtin functionality. As a fan and user of Rainbow, I'd like to ask the core team to give some insight into why they've chosen to tie directly into Eth? Why should implementing other tokens requiring ditching core functionality? 

Let's have some discussion on why the current implementation is the best option for this project.

DanielSinclair:
@owieth Appreciate the contribution! These changes may be deprecated by additional work we are evaluating for a `Custom Profile` API that could simplify overrides for that half of the RainbowKit experience (as opposed to replacing all UI with `Custom ConnectButton`). Open to re-opening this in the future when that API is ready.

owieth:
@DanielSinclair Thanks for replying, can you make an estimate when this `Custom Profile API` is released?

johnsmithberlin:
In general, I decided to take the advice of @nickbytes  and found the modal window file, honestly I don’t think that I could dig through all this code and find all the necessary components in order to display another token, but maybe there are some instructions about what needs to be done? At least a list of steps, since creating a new modal would be weird considering there's already a great RainbowKit, so why not just change some values ​​instead of doing a "little more work"?

Thank to all of you.

DanielSinclair:
@johnsmithberlin The styling system could become unstable when injecting custom UI, so we want to make sure those issues are accounted for with a new API. Ideally devs could selectively override the `Profile` modal instead of the full RainbowKit UI.

PR #855, opened 2022-11-05
feat: export useModalState

gleuch:
Export useModalState from ModalContext so clients can do things based on modal status

### DIFF
--- a/packages/rainbowkit/src/index.ts
+++ b/packages/rainbowkit/src/index.ts
@@ -6,6 +6,7 @@ export {
   useAccountModal,
   useChainModal,
   useConnectModal,
+  useModalState,
 } from './components/RainbowKitProvider/ModalContext';
 export { useAddRecentTransaction } from './transactions/useAddRecentTransaction';
 export {


nickbytes:
Thanks for the PR @gleuch – I like the idea of it. Just curious, what are some common use cases ?

gleuch:
@nickbytes - UX effects, mainly changing background animations when modals are open.

kunjaltandel24:
@nickbytes There's one use case where a page needs to show connectModal if user is not connected to wallet but page need to be show and when user closes the modal it need to redirect user to back page or home page.

joshsny:
> @nickbytes There's one use case where a page needs to show connectModal if user is not connected to wallet but page need to be displayed and when user closes the modal it need to redirect user to back page or home page.

This is also our use case, pages which require authentication to view.

marcinciarka:
Bump, I could use this as well. It's a simple change which will make my life easier 🙏

DanielSinclair:
We'll probably move forward with this @gleuch! Can you add some docs to `site/` and a changeset?

DanielSinclair (changes requested):
We'll probably move forward with this @gleuch! Can you add some docs to `site/` and a patch changeset?

Kharabet:
I'd love to be able to use this hook as well! @gleuch, are you still on it?

Kharabet:
@DanielSinclair why do you protect access to many useful hooks btw? 
I just found that `useAuthenticationStatus` would be a very nice option to determine whether the user really passed all the stages of verification and only then I can assume them as authenticated user

DanielSinclair:
@Kharabet Generally a bad practice; best to focus on public API stability to reduce side effects of unintended use or regressions with internal refactors. You'll want to instead rely on:
- `next-auth` `SessionProvider`'s `useSession`: https://next-auth.js.org/getting-started/client
- or server-side props to respond to state: https://www.rainbowkit.com/docs/authentication

marcinciarka:
> @Kharabet Generally a bad practice; best to focus on public API stability to reduce side effects of unintended use or regressions with internal refactors. You'll want to instead rely on:
> 
> * `next-auth` `SessionProvider`'s `useSession`: [next-auth.js.org/getting-started/client](https://next-auth.js.org/getting-started/client)
> * or server-side props: [rainbowkit.com/docs/authentication](https://www.rainbowkit.com/docs/authentication)

None of these can be used to check whether a user closed the modal window.

Kharabet:
> @Kharabet Generally a bad practice; best to focus on public API stability to reduce side effects of unintended use or regressions with internal refactors. You'll want to instead rely on:
> 
> * `next-auth` `SessionProvider`'s `useSession`: https://next-auth.js.org/getting-started/client
> * or server-side props: https://www.rainbowkit.com/docs/authentication

I am not going to use neither next.js or next-auth.js :)
So what you proposed don't really help me.

Well, but from the other side there is really no mechanism to control the Modal. You can only trigger it to open, but not to close or track its state. These actions are pretty basic and it's the best way to make rainbowkit more flexible and customizable for integrators

nickbytes:
@gleuch @Kharabet @marcinciarka would the [custom ConnectButton](https://www.rainbowkit.com/docs/custom-connect-button) work for your use case?

With a custom ConnectButton, you'll be able to see authentication status (@joshsny) and whether the modal is open or not (along with other modals like the chain modal, etc etc).

Kharabet:
> @gleuch @Kharabet @marcinciarka would the [custom ConnectButton](https://www.rainbowkit.com/docs/custom-connect-button) work for your use case?
> 
> With a custom ConnectButton, you'll be able to see authentication status (@joshsny) and whether the modal is open or not (along with other modals like the chain modal, etc etc).


I will try it for sure. Didn't get to CustomButton yet. But I will let you know how it is going

eriktoor:
Does the author intend to land this?

Kharabet:
I made a workaround how to make these fields from Custom Connect Button to be accessible via redux. But I wish no one will see this code 🙈 

The best solution for sure is just to re-export these internal hooks on the package level. 
I understand your concern about Public API maintenance, but with these restrictions/incapsulation make the modals to be barely customizable and/or controllable

asghaier76:
@gleuch @DanielSinclair any update on this, this is really a needed feature since it will allow easily to check if the user proceeded with wallet connect or just dismissed the modal

DanielSinclair:
Opened a new PR #1251 to push this forward. Have opted to use the existing hooks instead of `useModalState()`. Let me know if this meets your needs.

```tsx
const { connectModalOpen } = useConnectModal();
const { accountModalOpen } = useAccountModal();
const { chainModalOpen } = useChainModal();
```

joshsny:
@DanielSinclair This looks good and meets the needs we had for this.

Kharabet:
> Opened a new PR #1251 to push this forward. Have opted to use the existing hooks instead of `useModalState()`. Let me know if this meets your needs.
> 
> ```tsx
> const { connectModalOpen } = useConnectModal();
> const { accountModalOpen } = useAccountModal();
> const { chainModalOpen } = useChainModal();
> ```
Yes, that helps a lot!

I also would really like to see exported `useAuthenticationStatus` hook 🙏

DanielSinclair:
@Kharabet Can have a look at that! Can you open a new issue related to your SIWE needs? We would like to improve the Authentication API and can discuss any improvements you would like to see.

PR #879, opened 2022-11-18
fix: updated chains available on wagmi chainlist

0xOptimism:
Updating the chains supported by WAGMI chain list and Rainbowkit.


nickbytes (approved):

nickbytes:
Thanks for the docs update @0xOptimism !

PR #895, opened 2022-11-30
Export AuthenticationContext

aalmada:
When using a custom button to signout, it's required a call to the `useAuthenticationAdapter` hook.

This currently results in the error:

```
Module not found: Error: Package path ./dist/components/RainbowKitProvider/AuthenticationContext is not exported from package /Users/.../node_modules/@rainbow-me/rainbowkit (see exports field in /Users/.../node_modules/@rainbow-me/rainbowkit/package.json)
```

This PR simply exports `AuthenticationContext`.

### DIFF
--- a/packages/rainbowkit/src/components/RainbowKitProvider/AuthenticationContext.tsx
+++ b/packages/rainbowkit/src/components/RainbowKitProvider/AuthenticationContext.tsx
@@ -37,7 +37,7 @@ export function createAuthenticationAdapter<Message>(
   return adapter;
 }
 
-const AuthenticationContext = createContext<AuthenticationConfig<any> | null>(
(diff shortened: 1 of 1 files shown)


nickbytes:
@aalmada do you have a minimal reproducible example of this?

aj-may:
Doesn't calling disconnect from the wagmi hook automatically log out the user?  Looks like it does here:
https://github.com/rainbow-me/rainbowkit/blob/737a1d6ee13f409fc2060fa55d6c741a7fa574db/packages/rainbowkit/src/components/RainbowKitProvider/AuthenticationContext.tsx#L58

DanielSinclair:
@aj-may is correct. `useDisconnect` via Wagmi would be the appropriate implementation here. Here are the docs for reference: https://wagmi.sh/react/hooks/useDisconnect

PR #921, opened 2022-12-20
feat: add zerionWallet

zerts:


packages/rainbowkit/src/wallets/walletConnectors/zerionWallet/zerionWallet.ts
```diff
@@ -0,0 +1,118 @@
...
+
+  return {
+    id: 'zerion',
+    name: 'Zerion',
+    iconUrl: async () => (await import('./zerionWallet.svg')).default,
+    iconAccent: '#2962ef',
+    iconBackground: '#fff',
+    installed: !shouldUseWalletConnect ? isZerionInjected : undefined,
+    downloadUrls: {
+      browserExtension: '#',
```
DanielSinclair, line 34: Can we add a URL here, even if it's a redirect/placeholder? Example `https://zerion.io/download`. We will re-consider multiple browser store links in the near future, but we like that Phantom aggregates download options by linking to: `https://phantom.app/download`

DanielSinclair (changes requested):
Please revert changes to `getDefaultWallets`. We have a better solution in the works to resolve the default wallet list issue, and we'd like to solely focus on connector support and stability within RainbowKit.

zerts:
Hey, @DanielSinclair !
I've reverted changes to getDefaultWallets. Is it ok for now, or I should revert all calls of zerionWallet({ chains }) as well?

site/components/Provider/Provider.tsx
```diff
@@ -38,6 +39,7 @@ const connectors = connectorsForWallets([
       omniWallet({ chains }),
       imTokenWallet({ chains }),
       ledgerWallet({ chains }),
+      zerionWallet({ chains }),
```
DanielSinclair, line 42: Reverting this last addition to the docs site should wrap that part up; only because the extension is not yet available.

DanielSinclair (changes requested):

DanielSinclair:
@zerts Am seeing a few reliability problems on our example dApp deployment: https://rainbowkit-example-git-fork-zerts-add-zerio-6a8b08-rainbowdotme.vercel.app/
Other teams are bumping into an issue that appears to be related to provider injection delays; if RainbowKit doesn't see the flag at the time of connector initialization, it falls back to WalletConnect. We're not seeing this with MetaMask; likely because that is optimized.
Screenshot 2023-01-20 at 1 04 39 PM

everdimension:
@DanielSinclair 
The script that creates `window.ethereum` global is injected synchronously. You can check that `window.ethereum` is defined by adding an inline script to the head element of your test page, e.g.:
```
<head>
    <!-- other elements -->
    <script>
        console.log('ethereum provider is defined:', Boolean(window.ethereum));
    </script>
</head>
```

I did stumble upon inconsistencies in Rainbow kit trying to detect an Injected Wallet.

I didn't have a chance to look at how you do it, but if you need help trying to nail down the bug, feel free to reach out.

Btw, I wasn't able to reproduce the above bug on your test page, but I did see it in some dapps.

site/components/Provider/Provider.tsx
zerts: Ok!

examples/with-next/pages/_app.tsx
```diff
@@ -47,6 +48,7 @@ const connectors = connectorsForWallets([
       argentWallet({ chains }),
       trustWallet({ chains }),
       ledgerWallet({ chains }),
+      zerionWallet({ chains }),
```
DanielSinclair, line 51: Revert if possible

DanielSinclair (changes requested):

everdimension:
@DanielSinclair 
We've dug deeper into extension lifecycle issues and it turned out that due to what seems to be a chrome bug it was not guaranteed that the `ethereum` global was always injected before site scripts.

We've found a workaround that should provide a more robust behavior. We can send you the latest extension build.

DanielSinclair:
Looks like Wagmi didn't have a check for `isZerion` yet, so opened a PR to introduce that: https://github.com/wagmi-dev/references/pull/187
Cherry-picked this PR to #1108 to merge

PR #922, opened 2022-12-20
Remix.config for Wagmi

ertyurk:
Due to updates of Wagmi, I had to update serverDependencies to bundle ESM modules to make this example work

### DIFF
--- a/examples/with-remix/remix.config.js
+++ b/examples/with-remix/remix.config.js
@@ -10,7 +10,7 @@ module.exports = {
   serverDependenciesToBundle: [
     '@rainbow-me/rainbowkit',
     '@rainbow-me/rainbowkit/wallets',
-    'wagmi',
-    '@wagmi/core',
+    /^@?wagmi.*/,
+    'zustand/vanilla',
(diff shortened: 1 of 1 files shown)


examples/with-remix/remix.config.js
```diff
@@ -10,7 +10,7 @@ module.exports = {
   serverDependenciesToBundle: [
     '@rainbow-me/rainbowkit',
     '@rainbow-me/rainbowkit/wallets',
-    'wagmi',
-    '@wagmi/core',
+    /^@?wagmi.*/,
+    'zustand/vanilla',
```
nickbytes, line 14: @ertyurk what's reasoning for adding zustand here as well? I'm unfamiliar with some of the remix config.

ertyurk: Hi. afaik, zustand/vanilla is used at wagmi as well. without it, the Remix JS example did not work at that time.

DanielSinclair:
Should now be resolved with additional tweaks in #973!

PR #928, opened 2022-12-22
Update SIWE peer and dev dependencies

SSardorf:
Was getting `siwe.verify is not a function` before.

Updated to 2.0.3 to be in accordance with the latest version of SIWE


DanielSinclair:
Closing in favor of #961 while we wait for the settled `siwe` API coming in v3.

PR #938, opened 2023-01-05
feat: add Ledger Connect Kit, wip

hlopes-ledger:
Updates the Ledger connector to use Ledger Connect Kit, our library that allows using the Ledger Connect extension (currently Safari on iOS and macOS, Chrome is next) or fallback to WalletConnect.

Work in progress.

### DIFF
--- a/packages/rainbowkit/src/wallets/walletConnectors/ledgerWallet/ledgerWallet.ts
+++ b/packages/rainbowkit/src/wallets/walletConnectors/ledgerWallet/ledgerWallet.ts
@@ -1,42 +1,106 @@
 /* eslint-disable sort-keys-fix/sort-keys-fix */
+import { LedgerConnector } from '@wagmi/connectors/ledger';
+import { Connector } from 'wagmi';
 import { Chain } from '../../../components/RainbowKitProvider/RainbowKitChainContext';
-import { isAndroid } from '../../../utils/isMobile';
+import { isChrome, isIOS, isMacOS } from '../../../utils/isMobile';
 import { Wallet } from '../../Wallet';
-import { getWalletConnectConnector } from '../../getWalletConnectConnector';
 
 export interface LedgerWalletOptions {
   chains: Chain[];
+  enableDebugLogs?: boolean;
+  chainId?: number;
+  bridge?: string;
+  infuraId?: string;
+  rpc?: { [chainId: number]: string };
 }
 
-export const ledgerWallet = ({ chains }: LedgerWalletOptions): Wallet => ({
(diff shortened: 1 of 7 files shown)


packages/rainbowkit/src/wallets/walletConnectors/ledgerWallet/ledgerWallet.ts
```diff
@@ -1,42 +1,106 @@
...
+  rpc,
+}: LedgerWalletOptions): Wallet => {
+  // TODO check Connect support using hardcoded logic, will be out of sync with
+  // Connect Kit when it is updated, until DApps update to an updated version of
+  // RainbowKit
+  const isLedgerConnectEnabled =
+    typeof window !== 'undefined' && window.ethereum?.isLedgerConnect === true;
+  const ios = isIOS();
+  const macOS = isMacOS();
+  const isLedgerConnectSupported = (ios || macOS) && !isChrome();
```
hlopes-ledger, line 31: We're currently hardcoding the logic to determine if the user's platform supports Connect and if it is enabled.
Other two approaches would be to get the provider from the wagmi connector and check a property set by Connect Kit, or use Connect Kit directly, but both are async operations and I see no way to perform those currently.

e.g. using the wagmi provider

```
// TODO check Connect support through the wagmi provider
//
// - it is Connect if isLedgerConnect == true
// - it is TryConnect if isLedgerConnectSupported == true
// - it is WalletConnect if !!connector == true
//
// We would use this to decide what to show to the user, not sure if there are
// disadvantages of doing it here.
//
// const connector = new LedgerConnector({
//   chains,
//   options: {
//     enableDebugLogs,
//     isHeadless: true,
//     chainId,
//     bridge,
//     rpc,
//   },
// }) as unknown as Connector<any, any, any>;
//
// TODO can't call this as it's async
// const provider = connector.getProvider();
// const isLedgerConnectEnabled = !!provider.isLedgerConnect;
// const isLedgerConnectSupported = !!provider.isLedgerConnectSupported;
```

e.g. using Connect Kit

```
// TODO check connect support by using Connect Kit directly
// TODO can't call it as it's async
//
// const connectKit = await loadConnectKit();
// const connectKitSupport = connectKitPromise.checkSupport({
//   providerType: SupportedProviders.Ethereum,
//   chainId,
//   rpc,
// });
// const {
//   isLedgerConnectEnabled,
//   isLedgerConnectSupported
// } = connectKitSupport;
```

DanielSinclair:
Closing in favor of our rework #1085

PR #952, opened 2023-01-12
Remove unmaintained Alchemy provider from examples

george-gkas:
This PR removes all the instances of Alchemy providers from examples. Until now, each Alchemy provider was initialized using a hardcoded API key which eventually became obsolete, breaking the app functionality due to the inability to connect to the Alchemy network. 

Another more versatile solution, if we still want to support Alchemy in example apps, would be to allow users to define an Alchemy provider themselves through an environmental variable, e.g., `ALCHEMY=API_KEY`. If `ALCHEMY` is present, then the corresponding example will be initialized through Alchemy, else it will be initialized through public providers. This is just an idea and is currently out of the scope of the submitted work; `reainbowkit` team will decide the best action. 

Please also note that the example `with-next-mint-nft` is obsolete. The contracts are deployed on Rinkeby, which was [deprecated](https://twitter.com/etherscan/status/1569311894279958531) due to incompatibility with ETH 2.


DanielSinclair (approved):

nickbytes (approved):

nickbytes:
Thanks for doing this cleanup @george-gkas !!!

PR #960, opened 2023-01-16
Update installation.mdx

jossifelefteriadis:
We never import or use `ethers` lib so I removed them from the docs


DanielSinclair:
Ethers is actually an important peer dependency of RainbowKit leveraged within the provider and transaction logic:
https://github.com/rainbow-me/rainbowkit/blob/01f61336fcc252ef4475987b1b65d5dabee690c8/packages/rainbowkit/src/transactions/TransactionStoreContext.tsx#L1

PR #963, opened 2023-01-17
Set connect modal state to closed after sign-in to prevent open on disconnect

aj-may:
When using an authentication adapter with rainbow kit, the connect modal state is never set to closed after sign-in, it just doesn't have an effect because the model component itself returns `null`.  The issue is that once the user disconnects, the modal instantly re-appears.

This PR just calls the function to close the connect modal once the authentication `verify` function returns successfully.

### DIFF
--- a/packages/rainbowkit/src/components/SignIn/SignIn.tsx
+++ b/packages/rainbowkit/src/components/SignIn/SignIn.tsx
@@ -101,6 +101,7 @@ export function SignIn({ onClose }: { onClose: () => void }) {
         const verified = await authAdapter.verify({ message, signature });
 
         if (verified) {
+          onClose();
           return;
         } else {
           throw new Error();


aj-may:
@markdalgleish I see you are an active contributor here.  This one should be a pretty quick review 😂

I'd love to get your eyes on it if you have the time.

I also have a PR in with wagmi that I think might address some of the other modal issues that are seen in dev mode with react strict mode turned on: https://github.com/wagmi-dev/wagmi/pull/1679

aj-may:
Ok, so technically this isn't the issue, but I think it still makes sense to merged.
This is what is happening:
* When authentication is enabled and a user connects their wallet, all modals are closed except the connect modal so the user can sign the SIWE message.
* When the user successfully authenticates, the state changes to `authenticated` so the connect modal itself returns `null`, however the connect modal state is still technically open.
* When a user disconnects, all modals are set to closed.

So, while I do think it would be nice to technically set the connect modal state to closed after successful authentication, that is not the reason that in dev mode the connect modal reappears after disconnect.

That is happening because of this bug:
https://github.com/wagmi-dev/wagmi/pull/1679

DanielSinclair:
@aj-may I'm not seeing an issue after bumping to your changes in Wagmi `0.11.4`. Does this problem persist? If so, could you send over a screen recording of the behavior?

PR #965, opened 2023-01-17
Add optional onSignIn and onSignOut callbacks to next-auth adapter

aj-may:
Implemented the same callbacks in the auth adapter I wrote for [UseSIWE](https://github.com/random-bits-studio/use-siwe)
https://github.com/random-bits-studio/rainbowkit-use-siwe-auth

Related Discussion: https://github.com/rainbow-me/rainbowkit/discussions/904

### DIFF
--- a/packages/rainbowkit-siwe-next-auth/src/RainbowKitSiweNextAuthProvider.tsx
+++ b/packages/rainbowkit-siwe-next-auth/src/RainbowKitSiweNextAuthProvider.tsx
@@ -23,6 +23,8 @@ export type GetSiweMessageOptions = () => ConfigurableMessageOptions;
 interface RainbowKitSiweNextAuthProviderProps {
   enabled?: boolean;
   getSiweMessageOptions?: GetSiweMessageOptions;
+  onSignIn?: () => void;
+  onSignOut?: () => void;
   children: ReactNode;
 }
 
@@ -30,6 +32,8 @@ export function RainbowKitSiweNextAuthProvider({
   children,
   enabled,
   getSiweMessageOptions,
+  onSignIn,
+  onSignOut,
 }: RainbowKitSiweNextAuthProviderProps) {
   const { status } = useSession();
   const adapter = useMemo(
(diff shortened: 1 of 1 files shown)


DanielSinclair:
I lean more towards relying on the session state from the `SessionProvider` via `useSession` in `next-auth/react`. What do you think about updating the docs and `with-next-siwe-next-auth` instead?

aj-may:
Not a bad idea, but it looks like callbacks are not currently fully supported in next-auth: https://github.com/nextauthjs/next-auth/blob/main/packages/next-auth/src/react/types.ts#L4

I am happy to add callbacks to my [UseSIWE](https://github.com/random-bits-studio/use-siwe) library though. (Would y'all be open to me adding UseSIWE as an authentication option in the docs?)

I can eventually see what the effort would be to contribute callbacks to next-auth, but I'm not a fan of their codebase...

DanielSinclair:
@aj-may The `useSession` hook should mitigate the need for callbacks as so: 
```js
const { status } = useSession()
```

DanielSinclair:
@aj-may What's the TL;DR on UseSIWE? Happy to add look into adding the adapter to the docs if there is a valuable feature set difference

aj-may:
@DanielSinclair so the returned `status` could be used along with `useEffect` and some other boilerplate code to get it to work as needed.  Having a callback would be a much better developer experience.

I tried to make a PR but no luck. haha
https://github.com/nextauthjs/next-auth/pull/6574

I agree it would be better implemented as part of the `useSession` hook upstream, but given there's no appetite to add features at this time, this might be the next best option. 

[UseSIWE](https://github.com/random-bits-studio/use-siwe) Is a Sign-In with Ethereum library I wrote because I was super frustrated with how hacky the implementation of SIWE felt with next-auth.

It's a much simpler library overall given it doesn't support other authentication methods, which generally are not desired by Web3 app developers.

I'm super happy with how it came out, and it even comes with a [RainbowKit Authentication Adapter](https://github.com/random-bits-studio/rainbowkit-use-siwe-auth) 🌈

Love to get your feedback on it if you have a chance to take a look 😊

aj-may:
@DanielSinclair Here's an example of what a developer might need to do to get the same functionality without callback support in either `next-auth` or the adapter: https://github.com/aj-may/rainbowkit/pull/1/files

PR #1003, opened 2023-02-03
feat: walletconnect v2 support

DanielSinclair:
- Added `version` and `providerId` to WalletConnector connector API
- Defaulting connectors to v1
- Bumped WalletConnect branded connector to v2
- Added `WALLETCONNECT_PROVIDER_ID` and `NEXT_PUBLIC_WALLETCONNECT_PROVIDER_ID` envs

### DIFF
--- a/packages/rainbowkit/src/wallets/getWalletConnectConnector.ts
+++ b/packages/rainbowkit/src/wallets/getWalletConnectConnector.ts
@@ -15,15 +15,24 @@ function createConnector(options: WalletConnectConnectorOptions) {
 
 export function getWalletConnectConnector({
   chains,
+  projectId,
   qrcode = false,
+  version = '1',
 }: {
   chains: Chain[];
   qrcode?: boolean;
+  version?: '1' | '2';
+  projectId?: string;
 }) {
   const options: WalletConnectConnectorOptions = {
     chains,
     options: {
+      projectId:
+        projectId ||
+        (process.env.WALLETCONNECT_PROJECT_ID as string) ||
+        (process.env.NEXT_PUBLIC_WALLETCONNECT_PROJECT_ID as string),
       qrcode,
+      version,
     },
   };
 
--- a/examples/with-create-react-app/package.json
+++ b/examples/with-create-react-app/package.json
@@ -16,7 +16,7 @@
     "react": "^18.1.0",
     "typescript": "^4.9.4",
     "util": "0.12.4",
-    "wagmi": "^0.11.0",
+    "wagmi": "^0.11.3",
     "web-vitals": "^2.1.4",
     "buffer": "npm:buffer@6.0.3"
   },
--- a/examples/with-create-react-app/package.json
+++ b/examples/with-create-react-app/package.json
@@ -14,7 +14,7 @@
     "react-dom": "^18.1.0",
     "react-scripts": "5.0.1",
     "react": "^18.1.0",
-    "typescript": "^4.7.2",
+    "typescript": "^4.9.4",
     "util": "0.12.4",
     "wagmi": "^0.9.0",
     "web-vitals": "^2.1.4",
--- a/examples/with-create-react-app/package.json
+++ b/examples/with-create-react-app/package.json
@@ -16,7 +16,7 @@
     "react": "^18.1.0",
     "typescript": "^4.9.4",
     "util": "0.12.4",
-    "wagmi": "^0.9.0",
+    "wagmi": "^0.11.0",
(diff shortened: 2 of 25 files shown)


packages/rainbowkit/src/wallets/getWalletConnectConnector.ts
```diff
@@ -15,15 +15,24 @@ function createConnector(options: WalletConnectConnectorOptions) {
...
   qrcode?: boolean;
+  version?: '1' | '2';
+  projectId?: string;
 }) {
   const options: WalletConnectConnectorOptions = {
     chains,
     options: {
+      projectId:
+        projectId ||
+        (process.env.WALLETCONNECT_PROJECT_ID as string) ||
```
jxom, line 32: FYI, this might break in environments that don't inject `process.env` vars (Vite).

0xAsimetriq:
You'll likely need to also update which events you listen to for `display_uri` and `connect`, within web3modal we set these up as follows: https://github.com/WalletConnect/web3modal/blob/V2/chains/ethereum/src/client.ts#L57-L75 

Listening on provider itself is suggested, as you can utilise `.once`

fodisi:
Hey, I just wanted to check if rainbowkit depends on this PR to start supporting WalletConnect v2, or if this is just some sort of improvement. Just wondering about the impact of WalletConnect sunsetting v1 to rainbowkit.
Thanks

DanielSinclair:
@fodisi WalletConnect v2 is not currently supported. There are bugs and differences in the underlying UniversalConnector that we are working through with WalletConnect

fodisi:
> @fodisi WalletConnect v2 is not currently supported. There are bugs and differences in the underlying UniversalConnector that we are working through with WalletConnect

thanks a lot for the quick reply, @DanielSinclair !

adrianotadao:
@fodisi @DanielSinclair hey folks, sorry if this is a very silly question but there is something that is not that clear to me. 

> Just wondering about the impact of WalletConnect sunsetting v1 to rainbowkit and if the current 0.8.1 already has support for WC v2.

Will it affect, for example, Metamask wallet? I'm asking because I've seen some internal code that uses walletconnect for mobile etc etc.

DanielSinclair:
@adrianotadao Yes, each wallet will be affected and will need to migrate to WalletConnect v2. For RainbowKit, we will migrate each of these wallet connectors to v2 as the support becomes available. MetaMask currently hasn't released support, and other wallets are facing migration issues which we hope will soon be resolved. We will also leave a 'version` prop that allows custom connectors to continue to use v1 where necessary (like with custom bridge servers). dApps will not need to do anything except bump to the latest RainbowKit and Wagmi versions when support is finalized. We will release patch releases as each wallet connector rolls over to v2.

adrianotadao:
Hey @DanielSinclair, thank you for answering my question! I still have more questions =D

> MetaMask currently hasn't released support, and other wallets are facing migration issues which we hope will soon be resolved.

Does it mean that my app that relies on Metamask connection will break and I won't have any authentication system at all? :scream:

DanielSinclair:
@adrianotadao Correct. But, those teams are also working with WalletConnect, so I would assume WalletConnect won't deprecate v1 until they have migrated.

obernardovieira:
Would it be possible to have a beta version of this? Would love to test.

DanielSinclair:
@obernardovieira The WalletConnect team is still working on getting fixes into Wagmi, but we'll update this PR when those changes are settled.

fodisi:
Hi @DanielSinclair 
My latest understanding is that in order to have support for wallet connect v2, dApps need to upgrade wagmi to version 1.0.x (along w/ wagmi and other dependencies, + addressing potential breaking changes). Is this correct or are there other blockes/actions needed? Thanks a lot

DanielSinclair:
@fodisi We will start to migrate a set of wallets to WalletConnect v2 in a patch release very soon, and will complete migrating all RainbowKit connectors by June 14th. dApps will need to update RainbowKit on this timeframe. The RainbowKit 0.12.x API is already WalletConnect v2 compatible, but the wallet connectors have not yet been migrated. It is recommended that you begin to migrate your dApp to RainbowKit v1 and Wagmi v1, but we will also cherry-pick full WalletConnect v2 support as RainbowKit 0.12.15 on June 14th to make it easier to adopt for dApps that have not yet migrated.

fodisi:
@DanielSinclair Thanks a lot for all the info about the migration plan, schedule and references 🙏

Issue #836, opened 2022-10-21
[bug] Account Info modal opens after reconnect and cannot be dismissed


rtman:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### RainbowKit Version

0.7.3

### wagmi Version

0.7.5

### Current Behavior

When the dapp is disconnected from metamask, I click the connect wallet button and follow the instructions to connect my wallet. Once it is is connected I don't see a modal popup for account info. Thats good.

If I then disconnect again (from within the dapp not from metamask) and reconnect, the account info modal pops up and I cannot close it.

### Expected Behavior

I disconnect from the connected wallet and then reconnect to see no modal popup (or if there must be one, it is closable).

### Steps To Reproduce

1. Connect a wallet using rainbow kit and metamask
2. disconnect via rainbow kit
3. Reconnect via rainbow kit
4. See that account info modal pops up and is not closable.

### Link to Minimal Reproducible Example (CodeSandbox, StackBlitz, etc.)

I've so far been unable to reproduce this in codesandbox, still trying however.

### Anything else?

![Screen Shot 2022-10-21 at 3 30 57 PM](https://user-images.githubusercontent.com/20849135/197297892-3a4d10e1-1904-4287-a07f-995ad938d6c6.png)

0xAsynchronous:
+1 having the same issue

tamaka365:
use `wagmi 0.6.8` or lower version

tamaka365:
use `wagmi 0.6.8` or lower version

rtman:
@tamaka365 that change actually does work, not sure why. I'd really rather not use an old version of wagmi tho. Can any contributor comment on why this is may be the case?

mathesond2:
+1 same issue

rtman:
@mathesond2 @0xAsynchronous are you able to repro this in a codesandbox? I've tried but haven't been able to yet, or at least I haven't without implementing more of our app logic.

mathesond2:
@rtman thanks for getting back to me, I've created a [Loom video recording](https://www.loom.com/share/020c680054f545298e8778e02b310c98) to illustrate this issue along with an [associated small repo](https://github.com/mathesond2/rainbowkit-issue) available for checkout. Please let me know if I can help further, happy to do so :)

mathesond2:
Just a heads up: this should also be linked with #807

jacobedawson:
+1 having this issue with rainbowkit 0.7.8

[...]

Issue #837, opened 2022-10-22
[bug] <the modal cannot close>


wwei-github:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### RainbowKit Version

0.7.2

### wagmi Version

0.7.5

### Current Behavior

Using the official example, click connectbutton to connect to the matemask wallet, and when it's done, click disconnect, and the X in the popover won't close the popover
 

### Expected Behavior

I hope it can cancel the modal

### Steps To Reproduce

click connectButton -> matemask -> disconnect -> click modal X

rtman:
Also encounter this occasionally.

mathesond2:
+1 I've got the same issue currently, happy to speak more to it

jacobedawson:
Seems to be related to this error - there's some issue with changing the state on the unmounted modals:

![image](https://user-images.githubusercontent.com/4788506/198268991-90c9c42d-dedd-4ad2-b06c-5c3557d99f4a.png)

actuallymentor:
Can confirm the same issue, also same error as @jacobedawson.

cattingcat:
I have the same issue, but I have error log:
```
react-remove-scroll-bar: cannot calculate scrollbar size because it is removed (overflow:hidden on body 
```

My hypothesis is that:
When you open connect-wallet dialog - scroll on body really becomes hidden, 
After that you are clicking connect with metamask
Old dialog closes and new dialog with wallet details opens, and it is trying to hide scrollbar again, but library for hiding wallet throws exception because scrollbar already hidden

https://github.com/theKashey/react-remove-scroll-bar/blob/3dd80e28c92b8aac025e41f4258ff926cdc88af9/src/utils.ts#L22

Tingwei-Justin:
Same issue. Hope can get a quick fix

wassienabe:
Same issue here, not seeing any errors however, the close button simply doesn't respond after disconnecting elsewhere.

outdoteth:
+1. For me, the "disconnect" and "copy address" buttons are responsive, it's only the "x" that doesnt respond.

jacobedawson:
@wassienabe @outdoteth @Tingwei-Justin @cattingcat @actuallymentor found the issue / quick fix: https://github.com/rainbow-me/rainbowkit/issues/836#issuecomment-1303571558

sarinniural:
I am seeing the same issue.

dujomaton:
[...]

PR #1004, opened 2023-02-06
fix: close modal state on dialog unmount

johnrdoty92:
## Changes
- Added a call to `onClose` in the `useEffect` clean up return function in `Dialog.tsx`

## Fixes Issues
#836
#837

Based on what I've read, the issue exists only in development mode with React Strict Mode enabled. Rather than remove strict mode,  I think it makes sense to do some clean up on the dialog state. For example, in the `AccountModal`, if no `address` is present, the component returns `null`, but it is possible that the modal's state is still set to `true`. Below are two screen recordings taken from the vite example, before and after the fix.

https://user-images.githubusercontent.com/82186307/216893941-0cdcc1ba-b5fb-4051-b83e-1091b46d26b1.mov

https://user-images.githubusercontent.com/82186307/216894023-0eae670e-0057-46e5-ab59-706f14ea1f9e.mov

### DIFF
--- a/packages/rainbowkit/src/components/Dialog/Dialog.tsx
+++ b/packages/rainbowkit/src/components/Dialog/Dialog.tsx
@@ -39,6 +39,7 @@ export function Dialog({ children, onClose, open, titleId }: DialogProps) {
     setBodyScrollable(
       getComputedStyle(window.document.body).overflow !== 'hidden'
     );
+    return () => onClose();
   }, []);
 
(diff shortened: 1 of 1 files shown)


DanielSinclair:
The Wagmi `0.11.4` release corrected the core issue. Closing in favor of adopting that release: https://github.com/rainbow-me/rainbowkit/issues/837#issuecomment-1422165371

PR #1015, opened 2023-02-13
fix: injected wallet check for rainbow

estebanmino:
Adding this check to not show the "injected wallet" option while using the ~~extension~~

### DIFF
--- a/packages/rainbowkit/src/wallets/walletConnectors/injectedWallet/injectedWallet.ts
+++ b/packages/rainbowkit/src/wallets/walletConnectors/injectedWallet/injectedWallet.ts
@@ -21,7 +21,8 @@ export const injectedWallet = ({
       wallet =>
         wallet.installed &&
         (wallet.connector instanceof InjectedConnector ||
-          wallet.id === 'coinbase')
+          wallet.id === 'coinbase' ||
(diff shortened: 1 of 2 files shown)


nicoglennon (approved):

packages/rainbowkit/src/wallets/walletConnectors/injectedWallet/injectedWallet.ts
```diff
@@ -21,7 +21,8 @@ export const injectedWallet = ({
       wallet =>
         wallet.installed &&
         (wallet.connector instanceof InjectedConnector ||
```
DanielSinclair, line 23: This should catch the Rainbow connector, so we'll have to make changes to the `installed` logic in that connector instead

DanielSinclair (changes requested):
Seeing the same behavior on localhost after this change. 
Screenshot 2023-02-13 at 3 43 36 PM

packages/rainbowkit/src/wallets/walletConnectors/injectedWallet/injectedWallet.ts
estebanmino: yeah just pushed missed the param

PR #1029, opened 2023-02-16
feat: bitski extension support

pzuraq:
Adds support for Bitski's new browser extension wallet, which is currently in Beta. Note that there are some type hacks around the fact that the Ethereum types don't have `isBitski` on them, I have a PR to Wagmi open to fix that here: https://github.com/wagmi-dev/references/pull/116


packages/rainbowkit/src/wallets/getDefaultWallets.ts
```diff
@@ -28,6 +29,7 @@ export const getDefaultWallets = ({
         metaMaskWallet({ chains }),
         walletConnectWallet({ chains }),
         braveWallet({ chains }),
+        bitskiWallet({ chains }),
```
DanielSinclair, line 32: Revert addition here. Add to `packages/example` implementation for QA test deployment.

packages/rainbowkit/src/wallets/walletConnectors/bitskiWallet/bitskiWallet.ts
```diff
@@ -0,0 +1,38 @@
...
+  chains,
+  shimDisconnect,
+}: BitskiWalletOptions): Wallet => {
+  const installed =
+    typeof window !== 'undefined' &&
+    typeof window.ethereum !== 'undefined' &&
+    ((window.ethereum as any).isBitski === true ||
+      window.ethereum.providers?.find((p: any) => p.isBitski === true));
+
+  return {
```
DanielSinclair, line 21: The `=> ({ })` syntax visible in `braveWallet.ts` is preferred.

DanielSinclair (changes requested):
Just a few small changes needed. Please also send over a beta zip or add me to the Trusted Tester list to field test and QA. `daniel@rainbow.me`

packages/rainbowkit/src/wallets/getDefaultWallets.ts
pzuraq: The reason we added this was because if it's not there by default, then the Injected Wallet is what shows instead. We were hoping it would work like Web3Onboard, where if the user has Bitski installed it would show up without needing to manually configured. Is that not the case?

DanielSinclair: @pzuraq That's the expected behavior currently. We plan on rolling out 2 things soon:
- alterations to the `injectedWallet` connector so that the metadata defaults to the connector for the actively injected wallets, rather than just using fallback metadata. We are also evaluating forcing this connector as a default in the `Custom Wallet List` scenario, where dApps could have historically not included `injectedWallet`. This should make RainbowKit work better with things like the Gnosis Safe browser, as well as preserving the goals of the `Custom Wallet List`.
- a new community wallet list project that RainbowKit inherits so that Rainbow isn't the only arbiter of wallet popularity

pzuraq: Ah, that makes perfect sense! Sounds good, I've updated the PR per your feedback so it should be ready to go now. Let us know if there's anything we can help out with for the updates to the `injectedWallet` connector!

packages/rainbowkit/src/wallets/walletConnectors/bitskiWallet/bitskiWallet.ts
chronicIntrovert: @DanielSinclair - any updates here?

pzuraq:
@DanielSinclair any other issues with this PR? We'd like to get it in sooner rather than later, happy to make any more changes that are necessary!

DanielSinclair:
@pzuraq Mostly looks good after testing, but there's an issue where `installed` will return true when Bitski is uninstalled but MetaMask is installed. Could you have another look at your detection logic?

pzuraq:
@DanielSinclair so I'm not seeing that locally when I uninstall Bitski and only have MetaMask. The logic currently returns `undefined` for installed, I pushed a change to coerce that into a `false` but it did not seem to change the behavior. `installed` should be false if it's not installed, right?

DanielSinclair:
@pzuraq The double negation did the trick. It's working in the latest deployment.

DanielSinclair (approved):

nickbytes (approved):

PR #1037, opened 2023-02-20
feat: rabbyWallet support

heisenberg-2077:


DanielSinclair:
After testing, am bumping into a few issues when multiple wallets are installed. I see that Rabby has a default flip; this alleviates some of the problems. But, we're hoping to be able to support multiple wallets simultaneously with all RainbowKit wallet connectors as this new wave of injected wallets rolls out. Has Rabby implemented `window.ethereum.providers`? Details on the approach could help us figure out what we can greenlight, or changes we need to make on our end.

The unhappy test path looks like:
- Enable both MetaMask and Rabby in Chrome
- Connect Rabby via RainbowKit
- Disconnect Rabby
- Connect MetaMask via RainbowKit
The last step fails and the Rabby account provider takes precedent, blocking the user from interfacing with MetaMask until disabling the extension or using the Rabby flip and refreshing the page.

DanielSinclair:
We'll move forward with the connector support, but would definitely like to see support for `window.ethereum.providers` or `window.evmproviders` in Rabby so that we can maintain the RainbowKit switching UX. Please do reach back out to amend in the future.

vvvvvv1vvvvvv:
> We'll move forward with the connector support, but would definitely like to see support for `window.ethereum.providers` or `window.evmproviders` in Rabby so that we can maintain the RainbowKit switching UX. Please do reach back out to amend in the future.

Sure, will discuss for this

nickbytes (approved):

PR #1042, opened 2023-02-21
feat: xdefiWallet support

loxator:
Adds [XDEFI](https://xdefi.io) wallet as an option

### DIFF
--- a/.changeset/short-hornets-clap.md
+++ b/.changeset/short-hornets-clap.md
@@ -1,0 +1,6 @@
+---
+'@rainbow-me/rainbowkit': minor
+'example': patch
+---
+
+Added support for XDEFI wallet
--- a/.changeset/short-hornets-clap.md
+++ b/.changeset/short-hornets-clap.md
@@ -1,6 +1,5 @@
 ---
 '@rainbow-me/rainbowkit': minor
-'example': patch
 ---
 
 Added support for XDEFI wallet
--- a/.changeset/short-hornets-clap.md
+++ b/.changeset/short-hornets-clap.md
@@ -2,4 +2,22 @@
 '@rainbow-me/rainbowkit': minor
 ---
 
-Added support for XDEFI wallet
+XDEFI Wallet Support
+
+**Example usage**
+
+```tsx
+import {
+  getDefaultWallets,
+  connectorsForWallets,
+} from '@rainbow-me/rainbowkit';
+import { xdefiWallet } from '@rainbow-me/rainbowkit/wallets';
+const { wallets } = getDefaultWallets({ appName, chains });
+const connectors = connectorsForWallets([
+  ...wallets,
+  {
+    groupName: 'Other',
+    wallets: [xdefiWallet({ chains })],
+  },
+]);
+```
--- a/packages/example/pages/_app.tsx
+++ b/packages/example/pages/_app.tsx
@@ -25,6 +25,7 @@ import {
   omniWallet,
   tahoWallet,
   trustWallet,
+  xdefiWallet,
   zerionWallet,
 } from '@rainbow-me/rainbowkit/wallets';
 
@@ -95,6 +96,7 @@ const connectors = connectorsForWallets([
       omniWallet({ chains, projectId }),
       tahoWallet({ chains }),
       trustWallet({ chains, projectId }),
+      xdefiWallet({ chains }),
       zerionWallet({ chains, projectId }),
     ],
   },
--- a/packages/rainbowkit/src/wallets/Wallet.ts
+++ b/packages/rainbowkit/src/wallets/Wallet.ts
@@ -39,7 +39,7 @@ export type Wallet<C extends Connector = Connector> = {
   iconUrl: string | (() => Promise<string>);
   iconAccent?: string;
   iconBackground: string;
-  installed?: boolean | (() => boolean | undefined) | Promise<boolean>;
+  installed?: boolean;
   downloadUrls?: {
     android?: string;
     ios?: string;
@@ -50,7 +50,7 @@ export type Wallet<C extends Connector = Connector> = {
     wallets: {
       id: string;
       connector: Connector;
-      installed?: boolean | (() => boolean | undefined) | Promise<boolean>;
+      installed?: boolean;
       name: string;
     }[];
   }) => boolean;
@@ -59,7 +59,7 @@ export type Wallet<C extends Connector = Connector> = {
 
 export type WalletList = { groupName: string; wallets: Wallet[] }[];
 
-export type WalletInstance = Omit<Wallet, 'createConnector'> &
+export type WalletInstance = Omit<Wallet, 'createConnector' | 'hidden'> &
   ReturnType<Wallet['createConnector']> & {
     index: number;
     groupIndex: number;
--- a/packages/rainbowkit/src/wallets/Wallet.ts
+++ b/packages/rainbowkit/src/wallets/Wallet.ts
@@ -39,7 +39,7 @@ export type Wallet<C extends Connector = Connector> = {
   iconUrl: string | (() => Promise<string>);
   iconAccent?: string;
   iconBackground: string;
-  installed?: boolean;
+  installed?: boolean | (() => boolean | undefined) | Promise<boolean>;
   downloadUrls?: {
     android?: string;
     ios?: string;
@@ -50,7 +50,7 @@ export type Wallet<C extends Connector = Connector> = {
     wallets: {
       id: string;
       connector: Connector;
-      installed?: boolean;
+      installed?: boolean | (() => boolean | undefined) | Promise<boolean>;
       name: string;
     }[];
   }) => boolean;
@@ -59,7 +59,7 @@ export type Wallet<C extends Connector = Connector> = {
 
 export type WalletList = { groupName: string; wallets: Wallet[] }[];
 
-export type WalletInstance = Omit<Wallet, 'createConnector' | 'hidden'> &
+export type WalletInstance = Omit<Wallet, 'createConnector'> &
   ReturnType<Wallet['createConnector']> & {
     index: number;
     groupIndex: number;
--- a/packages/rainbowkit/src/wallets/connectorsForWallets.ts
+++ b/packages/rainbowkit/src/wallets/connectorsForWallets.ts
@@ -118,6 +118,7 @@ export const connectorsForWallets = (walletList: WalletList) => {
           connector,
           groupIndex,
           groupName,
+          hidden,
           index,
           walletConnectModalConnector,
           ...walletMeta,
--- a/packages/rainbowkit/src/wallets/connectorsForWallets.ts
+++ b/packages/rainbowkit/src/wallets/connectorsForWallets.ts
@@ -118,7 +118,6 @@ export const connectorsForWallets = (walletList: WalletList) => {
           connector,
           groupIndex,
           groupName,
-          hidden,
           index,
           walletConnectModalConnector,
           ...walletMeta,
--- a/packages/rainbowkit/src/wallets/useWalletConnectors.ts
+++ b/packages/rainbowkit/src/wallets/useWalletConnectors.ts
@@ -1,3 +1,4 @@
+import { useEffect, useState } from 'react';
 import { Connector, useConnect } from 'wagmi';
 import { flatten } from '../utils/flatten';
 import { indexBy } from '../utils/indexBy';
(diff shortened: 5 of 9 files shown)


DanielSinclair:
Hey @loxator. Can you swap out the SVG with a non-transparent background?

DanielSinclair:
I've deployed the `rainbowkit-example` here, so please QA with this web deployment. Many teams have bumped into different behavior between localhost and a production environment because of injector limitations.

packages/rainbowkit/src/wallets/walletConnectors/xdefiWallet/xdefiWallet.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  return {
+    id: 'xdefi',
+    name: 'XDEFI Wallet',
+    installed: !!installed,
+    iconUrl: async () => (await import('./xdefiWallet.svg')).default,
+    iconBackground: '#fff',
+    downloadUrls: {
+      browserExtension:
+        'https://install.xdefi.io/?utm_source=rainbowKit&utm_medium=organic&utm_campaign=xdefi.io&utm_id=xdefi.io',
```
DanielSinclair, line 29: Please swap this out for a direct Chrome Web Store link. We'll shortly be introducing an improved schema to allow multiple store links to properly route users.

DanielSinclair (changes requested):

loxator:
@DanielSinclair , Thanks a lot for giving this a test run, really appreciate it! 

I have updated the SVG as per your comment. I hope that one works.

The issue you mentioned with the injector is actually a bit of custom logic on our end. We had issues with some dApps that would keep sending `eth_requestAccounts` continuously, even after the injector would reply with valid accounts. So we have a 2 second wait where by if the same `origin` requests the same `msg`, we tend to ignore it. 

If you wait for 2 seconds and click on `Retry` , it will work as expected. Please let me know if this is ok, or if its a cause for some regression, I will check it out further and discuss with my team to remove this logic if need be, as we would very much like to get integrated into RainbowKit. 

Thanks again!

DanielSinclair:
@loxator Got it. That logic should be okay. But I'm still bumping into an issue where the connector is not always detected; RainbowKit looks at the window on-load to do this.

https://user-images.githubusercontent.com/4412473/221380476-76491018-bb86-4bab-a654-2ae4e1ac5e53.mov

loxator:
@DanielSinclair , that's a nice catch! Did you have any other wallet enabled that overwrites `window.ethereum` object? I couldn't reproduce it when I had just XDEFI enabled, but with MetaMask enabled as well, it seems like `window.ethereum` is overwritten and hence `isXDEFI` property is removed from the `ethereum` object.

My proposed solution is to check for our provider which takes the namespace `xfi`, so something like 

```ts
const installed =
    typeof window !== 'undefined' &&
    (typeof window?.xfi !== 'undefined' ||
      (typeof window.ethereum !== 'undefined' &&
        (window.ethereum as any).isXDEFI === true));
```

However, `window.xfi` isn't part of the global typings of and hence typescript throws an error. I was wondering if its ok to create a types.d.ts in the root folder and add the `xfi` property to extend the `Windows` object? Or if you have a better idea in mind? Thanks again!

DanielSinclair:
@loxator Nope, no other wallet active at that time. Make sure to test on a remote deployment; Chrome will have different injection behavior on localhost. The preferred approach is for the wallet to utilize `window.ethereum.providers` on overwrite and for `installed` to check both `window.ethereum` and `window.ethereum.providers`. As far as the `window`, feel free to just use a `@ts-ignore` for the time being. These window typings will probably get included into `wagmi` upstream at some point.

loxator:
Hi @DanielSinclair , I have tried to test on the site example after introducing the new changes, which is depending on the provider we inject `window.xfi` instead of `window.ethereum` as I believe anyone could completely overwrite the latter.

It seems to be working well, hope with this change, we can get it merged 🙌  As always, thanks for all your help and valuable feedback!

loxator:
Hi @DanielSinclair, any updates on this?

loxator:
Apologies for the constant pings @DanielSinclair , but it would be awesome if you could take a quick look on this one 🙏

DanielSinclair:
@loxator Can't seem to get the detection working
Screenshot 2023-03-09 at 9 53 08 PM

loxator:
@DanielSinclair Trying to get this to work, but still not entirely sure why its behaving like this. One thing I have found is that if I do `Empty cache & Hard refresh` on the remote deployment, then it manages to detect the provider. After that it doesn't, but I am unsure as to why 😕  

Our provider is injected as soon as the webpage is loading but somehow the `installed` is managing to read the data before that. We aren't doing any async work either before injecting. I am continuing to check but I was wondering if you had any idea regarding this 🤔

loxator:
@DanielSinclair , firstly, I apologize for the constant deploys as I had no other way of testing 🙈

 I just realized you have a draft PR open #1095 which could potentially fix this issue. I will wait for that to be merged first and come back to this. I hope I haven't caused a lot of mess on vercel 🙏

loxator:
@DanielSinclair I have picked the commit from #1095 and added it to my PR. Furthermore, i have also made `installed` to have a `Promise<boolean>` type, so I can trigger XDEFI's provider with a `setTimeOut` so the provider has time to inject.

It would be awesome if we could get this merged 😄

site/components/Provider/Provider.tsx
```diff
@@ -38,6 +39,7 @@ const connectors = connectorsForWallets([
       omniWallet({ chains }),
       imTokenWallet({ chains }),
       ledgerWallet({ chains }),
+      xdefiWallet({ chains }),
```
DanielSinclair, line 42: Please revert; only need example in `packages/example`

packages/rainbowkit/src/wallets/walletConnectors/xdefiWallet/xdefiWallet.ts
```diff
@@ -0,0 +1,56 @@
...
+
+declare global {
+  interface Window {
+    xfi: any;
+  }
+}
+
+export interface XDEFIWalletOptions {
+  chains: Chain[];
+  shimDisconnect?: boolean;
```
DanielSinclair, line 14: Should remove `shimDisconnect` here

packages/rainbowkit/src/wallets/walletConnectors/xdefiWallet/xdefiWallet.ts
```diff
@@ -0,0 +1,56 @@
...
+}
+
+export const xdefiWallet = ({
+  chains,
+  shimDisconnect,
```
DanielSinclair, line 19: Should pass through `InjectedConnectorOptions`. Reference other connector implementations.

.changeset/short-hornets-clap.md
```diff
@@ -0,0 +1,6 @@
+---
+'@rainbow-me/rainbowkit': minor
+'example': patch
```
DanielSinclair, line 3: `example` patch not needed. Please reference recent connector addition changesets formatting and example

packages/example/pages/_app.tsx
```diff
@@ -125,6 +126,7 @@ const connectors = connectorsForWallets([
       tahoWallet({ chains }),
       trustWallet({ chains }),
       zerionWallet({ chains }),
+      xdefiWallet({ chains }),
```
DanielSinclair, line 129: Please alphabetize

packages/rainbowkit/src/wallets/Wallet.ts
DanielSinclair, line 1: Please revert; will introduce with another PR.

packages/rainbowkit/src/wallets/connectorsForWallets.ts
DanielSinclair, line 1: Please revert.

packages/rainbowkit/src/wallets/useWalletConnectors.ts
DanielSinclair, line 1: Please revert.

packages/rainbowkit/src/wallets/walletConnectors/index.ts
```diff
@@ -36,4 +37,5 @@ export {
   okxWallet,
   walletConnectWallet,
   zerionWallet,
+  xdefiWallet,
```
DanielSinclair, line 40: Please alphabetize here.

packages/rainbowkit/src/wallets/walletConnectors/injectedWallet/injectedWallet.ts
DanielSinclair, line 1: Please revert.

packages/rainbowkit/src/wallets/walletConnectors/xdefiWallet/xdefiWallet.ts
```diff
@@ -0,0 +1,56 @@
...
+          getProvider: () => {
+            return new Promise(resolve => {
+              setTimeout(() => {
+                const res =
+                  typeof window !== 'undefined' &&
+                  typeof window?.xfi !== 'undefined'
+                    ? (window.xfi?.ethereum as any)
+                    : undefined;
+                resolve(res);
+              }, 0);
```
DanielSinclair, line 49: Is `setTimeout` necessary here? If so, please explain. Please also adopt async/await.

DanielSinclair (changes requested):

.changeset/short-hornets-clap.md
loxator: Sorry about that, I had updated the `example` to test out the vercel builds and added it while running `changeset` by mistake, reverted.

packages/rainbowkit/src/wallets/walletConnectors/xdefiWallet/xdefiWallet.ts
```diff
@@ -0,0 +1,54 @@
...
+      browserExtension:
+        'https://chrome.google.com/webstore/detail/xdefi-wallet/hmeobnfnfcmdkdcmlblgagmfpfboieaf',
+    },
+    createConnector: () => ({
+      connector: new InjectedConnector({
+        chains,
+        options: {
+          //@ts-ignore
+          getProvider: () => {
+            return new Promise(resolve => {
```
DanielSinclair, line 38: Wagmi connectors currently only support a synchronous `getProvider`

DanielSinclair (changes requested):

DanielSinclair:
@loxator We probably won't move forward with #1095 as it currently stands, as the installed/provider readiness mismatch could lead to unexpected side effects for dApps. We're planning a larger refactor here that would rebuild connectors in a way that is more compatible with Wagmi. Connectors are currently built on page load and have a one-time `ready` check that is set to `true` when `getProvider` returns a window provider. RainbowKit will mark a wallet as installed when both `ready` is true and `installed` (the window check within this connector) is not `undefined`. We have heard changes in Manifest v3 lead to differences in script execution and caching, which can cause issues in production for extensions (without the same behavior on localhost). All of the wallets that encountered this ended up resolving the issue within their injector, but I don't believe any are opensource or have shared the exact solution. Limiting asynchronous work within the injector inpage script seems to be the most important factor. There are separate situations where the extension client will have limited resources (like Safari Extensions) that exacerbates this problem.

loxator:
@DanielSinclair , thanks to your insight, I think I have finally solved this! We were adding the `xfi` object to the window object once we had setup the streams for communicating with the website. This was most likely happening after the `DOMContentLoaded` event, and thats why injection was slow.

I have fixed this on our end, and hopefully with the new build it should be good. I will ping you once we release the version of the wallet, for the time being it would be great if you could authorize the deploy with my latest changes. Thanks!

loxator:
@DanielSinclair , new version for XDEFI wallet is released (https://chrome.google.com/webstore/detail/xdefi-wallet/hmeobnfnfcmdkdcmlblgagmfpfboieaf?hl=en) . Should be good to test now 🙏

.changeset/short-hornets-clap.md
DanielSinclair, line 1: Can you add some example code here? You can reference a recent connector addition changeset here https://github.com/rainbow-me/rainbowkit/releases/tag/%40rainbow-me%2Frainbowkit%400.12.5

packages/rainbowkit/src/wallets/walletConnectors/xdefiWallet/xdefiWallet.svg
DanielSinclair, line 1: Please replace with an unrounded, minified SVG. Usually fix these upon release, but don't want to break the gradient.

DanielSinclair (changes requested):
Otherwise looks good. Will test and merge this week

loxator:
>Otherwise looks good. Will test and merge this week
Tearing up reading this ngl 😂, updated my PR! Thanks for your help :D

loxator:
Did you get a chance to test this out @DanielSinclair ?

DanielSinclair:
Closing in favor of #1175 to merge

punithbm:
> thanks to your insight, I think I have finally solved this! We were adding the `xfi` object to the window object once we had setup the streams for communicating with the website. This was most likely happening after the `DOMContentLoaded` event, and thats why injection was slow on the subsequent reloads.
> 
> I have fixed this on our end, and hopefully with the new build it should be good. I will ping you once we release the version of the wallet, for the time being it would be great if you could authorize the deploy with my latest changes. Thanks!

@loxator is't possible to provide any more information on how you were able to fix this at your end? I am also facing something similar issue and there might be other wallets as well, If you can give more details on the fix which was done at wallet. It will be really helpful. Thank you

loxator:
@punithbm Sorry, I can't help without much context. What exactly are you facing? The issue on my end was that our providers which talk with the dApps weren't being injected into the dApp websites on subsequent reloads (basically the cached website loaded faster than our injection as we were injecting into the `window` object once our extension was able to communicate over a port with the dApp)

punithbm:
@loxator Apologies for the delayed response. Thank you for providing the details. Our extension is MV3 based, and we have managed to discover some similar workarounds. Finger crossed that implementing these solutions will resolve the issue in our extension.

PR #1049, opened 2023-02-23
Export Rainbow Wallet Connector

herbievine:


DanielSinclair:
Hi @herbievine. What's the usecase/goal you're trying to accomplish?

herbievine:
Just typecasting a variable, as am using a 3rd party package which has some TS differences... 

There's no reason this type shouldn't be exported is there?

DanielSinclair:
Thanks for this contribution. Pulling this change into a broader type improvement for Wagmi v1 #1254

PR #1078, opened 2023-03-15
fix: set shimDisconnect to true for all injected providers

### LINKED ISSUE #1054: [bug] Even after using useDisconnect hook from wagmi, it is not possible to signout and rainbow kit signs back in on page refresh

### RainbowKit Version

0.11.0

### wagmi Version

0.9.6

### Current Behavior

Screenshot from 2023-02-26 23-48-03
Screenshot from 2023-02-26 23-47-52
Screenshot from 2023-02-26 23-47-45

Step 1: sign into wallet with rainbow kit
Step 2: sign out of wallet using disconnect button (removes sign in status and it seems like you are disconnected)
Step 3: [BUG] after refreshing you auto-sign back in

### Expected Behavior

After disconnecting you should stay disconnected. 

### Steps To Reproduce

Step 1: sign into wallet with rainbow kit
Step 2: sign out of wallet using disconnect button (removes sign in status and it seems like you are disconnected)
Step 3: [BUG] after refreshing you auto-sign back in

_No response_

### Anything else?

_No response_

TateB:
## problem

currently `shimDisconnect` is an option when creating any wallet with an associated `InjectedConnector`. having this as an option has the unintended consequence of auto-connecting *any* injected wallet, since there is no differentiation between the initialisation of them.

e.g. using the following wallet list

```ts
const wallets = [
      bitskiWallet({ chains }),
      metaMaskWallet({ chains, shimDisconnect: true }),
];
```

if you connect and disconnect from `MetaMask`, then refresh the page, the wallet will auto-connect. 

## solution

this pr removes `shimDisconnect` as an option wherever it was available to ensure a fallback auto-connect is never made.

## alternatives

an alternative solution could be to only allow `shimDisconnect` to be false if the `installed` property is true, but since the option creates an awkward experience anyway, i don't think that's necessary.

## fixed issues

fixes #1054, which is enabled by `getDefaultWallets` not passing `shimDisconnect: true` to each injected connector wallet.

### DIFF
--- a/packages/rainbowkit/src/wallets/walletConnectors/bitskiWallet/bitskiWallet.ts
+++ b/packages/rainbowkit/src/wallets/walletConnectors/bitskiWallet/bitskiWallet.ts
@@ -5,13 +5,9 @@ import { Wallet } from '../../Wallet';
 
 export interface BitskiWalletOptions {
   chains: Chain[];
-  shimDisconnect?: boolean;
 }
 
-export const bitskiWallet = ({
-  chains,
-  shimDisconnect,
-}: BitskiWalletOptions): Wallet => ({
+export const bitskiWallet = ({ chains }: BitskiWalletOptions): Wallet => ({
   id: 'bitski',
   name: 'Bitski',
   installed:
@@ -28,7 +24,6 @@ export const bitskiWallet = ({
   createConnector: () => ({
     connector: new InjectedConnector({
       chains,
-      options: { shimDisconnect },
     }),
   }),
 });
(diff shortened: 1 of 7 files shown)


DanielSinclair:
@TateB `shimDisconnect` actually already defaults to true upstream in Wagmi, so RainbowKit doesn't pass a default. We want `shimDisconnect` to be enabled for the behavior you're expecting, but also to provide the option to disable to dApps using the Custom Wallet List. It appears there is an issue in Wagmi that is causing instability around these disconnects.

TateB:
it's because the options object passes `shimDisconnect` as undefined if there is no value set, which overrides the `true` default via a spread operator in the connector initialiser

DanielSinclair:
@TateB Ah, great catch. Created another PR based on your findings that allows for pass-through options. Let me know what you think: https://github.com/rainbow-me/rainbowkit/pull/1080

TateB:
looks good to me

PR #1156, opened 2023-04-08
fix: only display provided chains in ChainModal

### LINKED ISSUE #1259: [bug] Error disconnecting Trust Wallet

### RainbowKit Version

1.0.0

### wagmi Version

1.0.1

### Current Behavior

Disconnecting a Trust wallet gives an error.

### Expected Behavior

No errors disconnecting

### Steps To Reproduce

connect a trust wallet.
disconnect a trust wallet.

Can be done directly from your website.

_No response_

### Anything else?

image

I have multiple providers installed, eg Trust Wallet, MetaMask, and Coinbase wallet.  I disabled their extensions and the error still occurs so I don't believe it's related.

main-eb61dab5f7771ce3.js:1 TypeError: o.removeListener is not a function
    at _app-62475ccf5efec63d.js:1:1176200
    at ei (framework-ca98db3857d7b0ec.js:1:93368)
    at ti (framework-ca98db3857d7b0ec.js:1:93564)
    at Ss (framework-ca98db3857d7b0ec.js:1:113664)
    at framework-ca98db3857d7b0ec.js:1:112695
    at ws (framework-ca98db3857d7b0ec.js:1:112761)
    at is (framework-ca98db3857d7b0ec.js:1:106480)
    at jl (framework-ca98db3857d7b0ec.js:1:46916)
    at framework-ca98db3857d7b0ec.js:1:103875

greg-schrammel:
I have a specific use case where the app is optimism only, but we provide a embedded [socket.tech](https://socket.tech/) bridge widget
I don't want to confuse users enabling all chains to be connected thru the modal, but they must be supported and only switch when bridge asks to

if I do this: 
```jsx
const { chains, provider, webSocketProvider } = configureChains(
  [optimism, optimismGoerli, mainnet, polygon, arbitrum],
  [publicProvider()]
);

...

<RainbowKitProvider chains={[optimism, optimismGoerli]}>
```
I expect to see this 
Screenshot 2023-04-08 at 18 20 26

but get this instead
Screenshot 2023-04-08 at 18 16 33

we could replace the `provideRainbowKitChains` in `useRainbowKitChains` with a map like
```ts
import { rainbowKitChainMetadata } from '...'

const chains = useRainbowKitChains()
// if the chain passed in the provider doesn't have metadata check if rainbowkit provides a default
const icons = chains.map(chain => chain.iconUrl || rainbowKitChainMetadata[chain.id]?.iconUrl) 
```

this refactor (that's not in this pr) would make the second screenshot have the chain logos


packages/rainbowkit/src/components/ChainModal/ChainModal.tsx
```diff
@@ -24,48 +18,20 @@ export interface ChainModalProps {
...
-      provider?.removeListener('chainChanged', stopSwitching);
-    };
-  }, [activeConnector, onClose, stopSwitching]);
-
-  useEffect(() => {
-    if (networkError && networkError.name === 'UserRejectedRequestError') {
-      stopSwitching();
-    }
-  }, [networkError, stopSwitching]);
-
```
greg-schrammel, line 68: I'm pretty sure this is the same as this
```ts
const { chains, pendingChainId, reset, switchNetwork } = useSwitchNetwork({
    onSettled: () => {
      reset(); // reset mutation variables (eg. pendingChainId)
      onClose();
    },
  });
```

didn't find any differences in my testing, but easy to revert if the useEffects where proposital

packages/rainbowkit/src/components/ChainModal/ChainModal.tsx
```diff
@@ -104,114 +70,124 @@ export function ChainModal({ onClose, open }: ChainModalProps) {
...
-                          display="flex"
-                          flexDirection="row"
-                          justifyContent="space-between"
-                        >
+                  let status = 'idle';
+                  if (chain.id === pendingChainId) status = 'switching';
+                  if (chain.id === activeChain?.id) status = 'current';
+
+                  const isCurrentChain = status === 'current';
+                  const switching = status === 'switching';
```
greg-schrammel, line 83: this lil "finite state machine" aims to solve this issue
Screenshot 2023-04-08 at 18 43 01

before:
https://user-images.githubusercontent.com/6232729/230744078-b69f0ee2-09bc-4cfc-a490-cdef78169735.mov

after:

https://user-images.githubusercontent.com/6232729/230744084-d8a432f3-62c2-46d9-a485-92458813ec2a.mov

check how in this pr implementation, the double state won't happen, but the "connect" text color is white (?) any idea why is that? (it's only white in the switching -> connected -> close, when opening again it's the right color)
Screenshot 2023-04-08 at 18 53 24

greg-schrammel: > check how in this pr implementation, the double state won't happen, but the "connect" text color is white (?) any idea why is that? (it's only white in the switching -> connected -> close, when opening again it's the right color)

found that's due to the `<MenuButton />` having a bg color transition, not sure how to "fix" that, or if it should be fixed

DanielSinclair:
Looking good 👌 
Would just need you to create patch changeset from `pnpm changeset` that explains the bug fix. This may also be a good first component unit test since we had missed the `chains` mismatch scenario. Think you can spin up a `@testing-library/react` setup and test?

greg-schrammel:
done! I'm not good with changeset messages, open to rephrasing
also first time setting up tests with wagmi, lmk what you think

DanielSinclair (approved):

PR #1189, opened 2023-04-25
feat: walletconnect web3modal options

DanielSinclair:
- `options` param on `walletConnectWallet`
- removed explicit `qrcode` param in `getWalletConnectConnector` in favor of `options`
- overload functions for v1 and v2 typed option support in `getWalletConnectConnector`
- doc site changes for new options and additional information about v2 transition

### DIFF
--- a/packages/rainbowkit/src/wallets/walletConnectors/walletConnectWallet/walletConnectWallet.ts
+++ b/packages/rainbowkit/src/wallets/walletConnectors/walletConnectWallet/walletConnectWallet.ts
@@ -21,9 +21,12 @@ export const walletConnectWallet = ({
     const ios = isIOS();
 
     const connector = getWalletConnectConnector({
-      projectId,
+      version: '1',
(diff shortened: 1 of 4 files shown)


nickbytes (approved):

packages/rainbowkit/src/wallets/walletConnectors/walletConnectWallet/walletConnectWallet.ts
```diff
@@ -21,9 +21,12 @@ export const walletConnectWallet = ({
     const ios = isIOS();
 
     const connector = getWalletConnectConnector({
-      projectId,
+      version: '1',
```
sunwrobert, line 24: I think we need to update the WalletConnectWalletOptions with the new `options` field, take that as a param in this `walletConnectWallet` fn, and then pass those additional options into this `getWalletConnectConnector` fn

PR #1202, opened 2023-04-28
feat: blocto wallet

sanyu1225:
When using RainbowKit, you can now easily manage your cryptocurrencies with [Blocto Wallet](https://docs.blocto.app/blocto-sdk/javascript-sdk/evm-sdk) ! We are excited to announce that we have updated our sample code and added [Blocto Wallet](https://docs.blocto.app/blocto-sdk/javascript-sdk/evm-sdk) support to RainbowKit. This update will provide our users with a better user experience as they can now use their favorite wallet to manage their cryptocurrencies.

We understand how important this feature is to our users, so we have dedicated ourselves to implementing it to make it easier for you to use RainbowKit and [Blocto Wallet](https://docs.blocto.app/blocto-sdk/javascript-sdk/evm-sdk).

## Screenshot/Gif
image
image


DanielSinclair:
This type of wallet connector is unfortunately not supported at this time. RainbowKit also cannot bundle SDKs.

PR #1206, opened 2023-05-02
Improve Ledger connector, check if Extension enabled

hlopes-ledger:
- checks whether the user's platform supports the Extension
- checks whether the first chain is supported by the Extension
- checks whether the extension is enabled
- falls back to WalletConnect if one of the above is false


hlopes-ledger:
Closing this one and improving #1207

PR #1207, opened 2023-05-02
Improve Ledger connector, check chain

hlopes-ledger:
- checks whether the user's platform supports the Extension
- checks whether the first chain is supported by the Extension
- falls back to WalletConnect if one of the above is false


hlopes-ledger:
Replaced by #1237

PR #1243, opened 2023-05-15
example: with-next-app

DanielSinclair:
- new example `examples/with-next-app`

### DIFF
--- a/examples/with-next-app/app/layout.tsx
+++ b/examples/with-next-app/app/layout.tsx
@@ -1,0 +1,72 @@
+'use client';
+
+import '../styles/global.css';
+import '@rainbow-me/rainbowkit/styles.css';
+import {
+  RainbowKitProvider,
+  getDefaultWallets,
+  connectorsForWallets,
+} from '@rainbow-me/rainbowkit';
+import {
+  argentWallet,
+  trustWallet,
+  ledgerWallet,
+} from '@rainbow-me/rainbowkit/wallets';
+import { configureChains, createClient, WagmiConfig } from 'wagmi';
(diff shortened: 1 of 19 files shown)


examples/with-next-app/app/page.tsx
```diff
@@ -0,0 +1,19 @@
+'use client';
```
jxom, line 1: fwiw, i would avoid client components at the root and try to get it as close to the leaves as possible so conumers can still leverage server components if they want.

two options:
- compose `ConnectButton` into a separate component tagged as a client component,
- [add the `banner` option to esbuild config](https://github.com/wagmi-dev/wagmi/blob/main/packages/react/tsup.config.ts#L8-L10)

examples/with-next-app/app/layout.tsx
```diff
@@ -0,0 +1,76 @@
+'use client';
```
jxom, line 1: also avoid `"use client";` here and encapsulate client dependants - good example of this is on create-wagmi: https://github.com/wagmi-dev/create-wagmi/tree/main/templates/next/rainbowkit/src/app

DanielSinclair: @jxom Great advice; have adopted the Wagmi `providers.tsx` approach.

examples/with-next-app/app/page.tsx
DanielSinclair: Ah, was wondering how to prevent esbuild from dropping that `use client;` tag. Thanks!

nickbytes (approved):

PR #1295, opened 2023-06-14
feat: coreWallet support

gergelylovas:
This PR is meant to add support for the Core mobile and browser extension wallets: https://core.app/


packages/rainbowkit/src/wallets/walletConnectors/coreWallet/coreWallet.svg
DanielSinclair, line 1: Please use a square SVG, minify, and resize to `width="28" height="28" viewBox="0 0 28 28" `

packages/rainbowkit/src/wallets/useWalletConnectors.ts
DanielSinclair, line 1: Please revert these changes; this sounds like an issue with the Core Wallet provider.

packages/example/pages/_app.tsx
```diff
@@ -91,6 +92,7 @@ const connectors = connectorsForWallets([
     wallets: [
       argentWallet({ chains, projectId }),
       bitskiWallet({ chains }),
+      coreWallet({ chains }),
```
DanielSinclair, line 95: Missing `projectId`

packages/rainbowkit/src/wallets/walletConnectors/coreWallet/coreWallet.ts
```diff
@@ -0,0 +1,148 @@
...
+        'https://chrome.google.com/webstore/detail/core-crypto-wallet-nft-ex/agoakfejjabomempkjlepdflaleeobhb',
+      browserExtension: 'https://extension.core.app/',
+    },
+    createConnector: () => {
+      const connector = shouldUseWalletConnect
+        ? getWalletConnectConnector({ projectId, chains })
+        : new InjectedConnector({
+            chains,
+            options: {
+              name: 'Core',
```
DanielSinclair, line 79: `name` should be inferred from Wagmi. The mismatch should be addressed in Wagmi first. https://github.com/wagmi-dev/references/blob/746df4f7f0b19dd8ca59e0cc7b6e53aa296bc767/packages/connectors/src/utils/getInjectedName.ts#LL8C6-L8C6

site/data/docs/custom-wallet-list.mdx
```diff
@@ -157,6 +157,17 @@ coinbaseWallet(options: {
...
 ```
 
+#### Core Wallet
+
+```tsx
+import { coreWallet } from '@rainbow-me/rainbowkit/wallets';
+
+coreWallet(options: {
+  appName: string;
+  chains: Chain[];
```
DanielSinclair, line 167: `appName` is incorrect. Missing `projectId`

DanielSinclair (changes requested):

packages/rainbowkit/src/wallets/useWalletConnectors.ts
gergelylovas: Thank you for the fast review! 
I've reverted this change to decouple this discussion from the connector PR. 
I do think this error handling is missing from here and I it's not really a bug in the provider. 

My reasoning:
EIP-1193 specifies this error as a valid error for cases when the provider is not connected to any chain: https://eips.ethereum.org/EIPS/eip-1193#provider-errors

So the question is, why is Core not connected to any chain when locked? 
This was an intentional decision being in alignment with EIP-1193. Core stores everything encrypted in storage while locked for security reasons, including the active chain it should be connected to. This means when the wallet is locked, the provider is not connected to any chains, resulting the `eth_chainId` RPC call rightfully throwing the 4900 error. 

What we could do on Core's end is making a temporary connection to a network, lets say Avalanche, and immediately switch chains when the wallet gets unlocked. In my oppinion, while it would work, it's not an optimal UX and the code in your hook would also work unoptimally since the `walletChainId` could be incorrect. Which would throw off your inital chain switching mechanism.

DanielSinclair (approved):

PR #1310, opened 2023-06-21
feat: configurable auth button label

andreivcodes:
The button label of the SignIn component can be configured in AppContext. The default value remains "Send message", as it currently is.

This PR comes as a follow up to this Twitter thread
https://twitter.com/paulofonseca__/status/1671535529513426944

Screenshot 2023-06-21 at 22 58 17

### DIFF
--- a/packages/example/pages/_app.tsx
+++ b/packages/example/pages/_app.tsx
@@ -116,6 +116,7 @@ const wagmiConfig = createConfig({
 
 const demoAppInfo = {
   appName: 'Rainbowkit Demo',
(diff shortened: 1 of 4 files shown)


DanielSinclair:
Apologies about the delayed response here. Thank you for this suggestion. I'll go ahead and merge the "Sign message" label that you recommended with #1581. Because we have introduced i18n support, we unfortunately need to limit the surface area of customizable button labels, so it will more straightforward if we adopt your original suggestion rather than introducing this API. Let me know if you still think there is a need for more customization.

PR #1342, opened 2023-06-29
Update custom-wallets.mdx

viral-sangani:
Updated `getUri()` function.
Since WalletConnect does not support V1, needed to make `getUri()` compatible with V2.

In V1, the way to get the URI was - 
```js
const { uri } = (await connector.getProvider()).connector;
```

But this does not work anymore, so need to update the docs 

```js
const provider = await connector.getProvider();
const uri = await new Promise<string>(resolve => provider.once('display_uri', resolve))
return uri
```

### DIFF
--- a/site/data/docs/custom-wallets.mdx
+++ b/site/data/docs/custom-wallets.mdx
@@ -238,14 +238,18 @@ export const rainbow = ({
       mobile: {
         getUri: async () => {
           const provider = await connector.getProvider();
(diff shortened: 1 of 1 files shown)


DanielSinclair:
@viral-sangani Thanks for spotting this.

DanielSinclair:
Just needs a `pnpm run lint:fix' to fix the CI error.

viral-sangani:
@DanielSinclair Updated the PR with lint fixes

nickbytes (approved):

PR #1350, opened 2023-06-30
feat: allow loading wallets on demand; can reduce first load by 80+ KB gzipped

llllvvuu:
resolves https://github.com/rainbow-me/rainbowkit/discussions/1349 (EDIT: just found another workaround, see discussion OP)

Demo of a lazy loading use case:

https://github.com/rainbow-me/rainbowkit/assets/5601392/3c5c2995-9798-4b37-a1ac-2030944db084

savings can be more than 80KB (which is just the `getDefaultWallets()` savings) if lazy loading more than the default wallets. although the big one I suspect is WalletConnect?

### DIFF
--- a/packages/rainbowkit/build.js
+++ b/packages/rainbowkit/build.js
@@ -65,10 +65,7 @@ const mainBuild = esbuild.build({
 
     // esbuild needs these additional entry points in order to support tree shaking while also supporting CSS
     ...(await getAllEntryPoints('src/themes')),
-
-    // The build output is cleaner when bundling all components into a single chunk
-    // This is done assuming that consumers use most of the components in the package, which is a reasonable assumption for now
-    './src/components/index.ts',
+    ...(await getAllEntryPoints('src/wallets')),
   ],
   outdir: 'dist',
   watch: isWatching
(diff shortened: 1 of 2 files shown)


DanielSinclair:
Hey @llllvvuu! Can you walk me through the changes here and provide some more details of what improvements you're working towards? We typically wouldn't want to load wallets on demand as Wagmi and the connector list is prepared at load-time, for features like auto-reconnect and shim disconnects. We also have logic baked into the RainbowKit connector build routine that decides when to show our fallback wallet for scenarios where a dApp dev didn't include a connector (i.e. MetaMask) in their wallet list, but MetaMask's injected provider is available a user would still need to connect.

llllvvuu:
> We typically wouldn't want to load wallets on demand as Wagmi and the connector list is prepared at load-time, for features like auto-reconnect and shim disconnects.

In our app, we can afford to have the auto-reconnect happen later, while there are other things like hydrating links, accordions/collapses, fetch from subgraph, REST, JSON-RPC, etc that we don't want to block on loading 100-200KB of gzipped javascript (it's not SSR for reasons)

Here's a lazy import demo: https://github.com/llllvvuu/wagmi-react-stale-closure/blob/main/pages/_app.tsx

We have a setup that works now though, so it's no longer blocking for us.

PR #1351, opened 2023-06-30
fix: default projectId for example and docs site

llllvvuu:
This is necessary to import a fork of `rainbowkit` in a project. (For `file:` syntax in `package.json` `dependencies`, it may not be enough; user may also have to nuke the `node_modules` and reinstall just `qrcode`)

Another option would be allowing empty string for the `projectId` (more aligned with the TypeScript typing)


DanielSinclair:
@llllvvuu Thanks! Will adopt this 👍

nickbytes (approved):

PR #1355, opened 2023-07-02
refactor: removed walletconnect v1 to decrease bundle size by 40kb

VladimirMikulic:
Now that WalletConnect v1 is shut down, WalletConnect v1 support is removed in this PR to decrease RainbowKit's overall bundle size by 40kb.

### DIFF
--- a/.changeset/ten-otters-bathe.md
+++ b/.changeset/ten-otters-bathe.md
@@ -1,0 +1,5 @@
+---
+'@rainbow-me/rainbowkit': major
+---
+
+WalletConnect v1 support has been removed to reduce bundle size by ~40kb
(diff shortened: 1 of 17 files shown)


DanielSinclair:
@VladimirMikulic Thanks for your contribution here Vlad. We plan to keep WalletConnect v1 support for some time as some wallets and dApps are still relying on self-hosted relay servers to keep v1 alive.

VladimirMikulic:
Thanks for reply @DanielSinclair. WalletConnect announced they expect self-hosted relay servers to migrate too.
What do you think about releasing major version of RainbowKit without WalletConnect legacy/v1? This way everyone gets performance boost while the handful (if any?) of dapps that still need WalletConnect v1 use RainbowKit v1.

PR #1366, opened 2023-07-05
fix: don't close connect modal on disconnect

llllvvuu:
Connect modal should only be closed by connecting.

It's rare for it to be open when a disconnect happens, since you'd expect a disconnect to happen from an already connected state. But it can happen, e.g. if you have a handler to open the connect modal on disconnect, which causes a race condition with the handler to close the connect modal on disconnect.


llllvvuu:
@DanielSinclair Have a chance to take a look at this one-liner? Thanks

DanielSinclair:
@llllvvuu Hey! Can you clarify what behavior you were seeing and were hoping to correct?

DanielSinclair:
@llllvvuu We talked to a few other dApp devs that use the Authentication flow and decided not to adopt this change because it would have side effects on the traditional UX. But, we will soon work on a `Lazy SIWE` hook API that will expose more capability so that you can request a SIWE signature with your own button on your logged-out routes.

PR #1420, opened 2023-07-19
example: with-next-app, remove `appDir` next config option

MrTanoshii:
As per https://nextjs.org/docs/app/api-reference/next-config-js/appDir , the `appDir` option is no longer needed.


DanielSinclair:
Thanks for spotting this and opening the PR 🫡

DanielSinclair (approved):

PR #1534, opened 2023-09-27
feat: I18n (Internalization)

0xmkh:
Japanese Example:

image


packages/rainbowkit/src/css/sprinkles.css.ts
```diff
@@ -136,6 +136,8 @@ const responsiveProperties = defineProperties({
   },
   defaultCondition: 'smallScreen',
   properties: {
+    overflowY: ['auto', 'scroll', 'hidden'],
+    flexWrap: ['wrap', 'nowrap'],
```
KosmosKey, line 140: Adding these so we can have better responsive UI like this:
Screenshot 2023-09-27 at 21 24 09

packages/example/pages/_app.tsx
```diff
@@ -98,41 +100,57 @@ const { wallets } = getDefaultWallets({
...
-      zerionWallet({ chains, projectId }),
-    ],
-  },
-]);
+// In this example, the user's locale is stored in local storage, allowing for language switching functionality.
+// If defined, we retrieve the locale value and typecast it as `Language`.
+export const localStorageLocale =
+  typeof localStorage !== 'undefined'
+    ? (localStorage.getItem('locale') as Language)
+    : null;
```
DanielSinclair, line 108: I think we can drop the need for localstorage in this example; we want to show passing down the locale to the RainbowKitProvider prop tree. dApps can implement that however they want, but we should consider:
1) the router locale
2) checking `navigator.language` for the user's preferred language (I don't think that is handled by the above?)
3) Altering the drop-down selection
The drop-down default can be set by 1 and 2, and we keep track of the current selection in useContext or useState; this will be a better pattern to demonstrate to developers

packages/rainbowkit/src/components/RainbowKitProvider/RainbowKitProvider.tsx
DanielSinclair, line 1: We should expose `locale` in `RainbowKitProviderProps` instead of `connectorsForWallets`. This will be a cleaner pattern to bubble down the locale into the prop tree and aligns with our existing options API (i.e. `modalSize`). Connectors should be rebuilt as little as possible as it has downstream effects on Wagmi, while the RainbowKit UI components can be rebuilt easily in the React tree. `locale` seems like the best term for this API prop as it specifies region + language

packages/rainbowkit/src/locales/index.ts
```diff
@@ -0,0 +1,77 @@
...
+  ES_419 = 'es_419',
+  FR_FR = 'fr_FR',
+  JA_JP = 'ja_JP',
+  PT_BR = 'pt_BR',
+  ZH_CN = 'zh_CN',
+  ID_ID = 'id_ID',
+  HI_IN = 'hi_IN',
+  TR_TR = 'tr_TR',
+  RU_RU = 'ru_RU',
+}
```
DanielSinclair, line 24: Let's rename this to `Locale` for language+region specificity

packages/rainbowkit/src/locales/index.ts
```diff
@@ -0,0 +1,77 @@
...
+    return;
+  }
+
+  const localStorageLocale = getLocalStorageLocale();
+  // Remove language from localStorage if none is specified and a language is already saved,
+  // to avoid potential language saved confusions.
+  if (localStorageLocale) {
+    localStorage.removeItem(storageKey);
+  }
+};
```
DanielSinclair, line 77: Because we will migrate to the options `locale` prop, are we able to deprecate the need for localstorage? This seems a bit error prone. The only exception outside of the React tree should be the connectors. Some ideas:
1) `import { i18n }` should always return the same instance of `i18n` so we can update `i18n.locale` at the time the option prop is passed into the provider, and rely on that to be updated when/if a connector is rebuilt; but that means we need to think more about connector refreshes
2) Can we instead return the string key from the connector and simply rely on the component update lifecycle and i18n instance in the React tree?

DanielSinclair (changes requested):
Added some ideas to simplify this a bit. Otherwise awesome progress 👌

packages/example/pages/_app.tsx
```diff
@@ -4,6 +4,7 @@ import './global.css';
 import {
   AvatarComponent,
   DisclaimerComponent,
+  Language,
```
DanielSinclair, line 7: I think we can remove the need for exporting Language; the type exposed in the RainbowKitProvider prop should be enough, maybe? There shouldn't be extensive casting needs.

packages/rainbowkit/src/components/ConnectOptions/DesktopOptions.tsx
```diff
@@ -248,7 +250,9 @@ export function DesktopOptions({ onClose }: { onClose: () => void }) {
           wallet={selectedWallet}
         />
       );
-      headerLabel = selectedWallet && `Get ${selectedWallet.name}`;
+      headerLabel =
+        selectedWallet &&
+        `${i18n.t('intro.wallet.scan.get')} ${selectedWallet.name}`;
```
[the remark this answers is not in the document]
KosmosKey: Good idea. Let's do `%{name}` like this 🙏

image

packages/example/pages/_app.tsx
KosmosKey: Yeah let's get rid of `localStorage` and focus on drop down and state managment.

All things added including `navigator.language` ✨

packages/rainbowkit/src/locales/index.ts
KosmosKey: Perfect ✨. I will use singleton rule to share one instance along with the application at run time 🙌.

I've also added key strings to the connector instructions so instead of importing it in every wallet the `connectorForWallet` function will make sure to use `i18n` singleton instance for those instructions.

packages/example/pages/_app.tsx
KosmosKey: Good catch. I have update on that. In most of typescript projects when using `enum` types the prop that requires the `enum` type will throw error if you don't import and specify it like this:

image

VS

image

KosmosKey: Some libraries like alchemy-sdk also provides `enum` types in their library to avoid type errors 👻

packages/rainbowkit/src/components/RainbowKitProvider/RainbowKitProvider.tsx
KosmosKey: Fully agree. `prop` would be the best way to update state. In this PR i will introduce refreshing connectors from wagmi and then make the `language` prop be fully accessible within the rainbowkit 🙏

packages/example/next.config.js
```diff
@@ -1,4 +1,19 @@
...
+      'ja_JP',
+      'pt_BR',
+      'zh_CN',
+      'id_ID',
+      'hi_IN',
+      'tr_TR',
+      'ru_RU',
+    ],
+  },
   reactStrictMode: true,
```
KosmosKey, line 18: These will be the `locales` we will support. People who use this feature will also have access to routing which will help `rainbowkit` stay in sync with any changes to the `locale` prop.

Example from next.js:

image

packages/example/pages/_app.tsx
```diff
@@ -246,9 +254,36 @@ function RainbowKitApp({
...
+    }
+
+    return locale;
+  };
+
+  // biome-ignore lint/nursery/useExhaustiveDependencies: TODO
+  useEffect(() => {
+    const userBrowserDetectedLocale = getUserDetectedBrowserLocale();
+    setLocale(userBrowserDetectedLocale);
+  }, []);
```
KosmosKey, line 285: Some `biome-ignore` for now. Will fix this later marked as `TODO`.

Here we will be detecting users `locale` from their browser using `navigator` 🙌

packages/example/utils/locale.ts
```diff
@@ -0,0 +1,51 @@
...
+  if (typeof window !== 'undefined' && typeof navigator !== 'undefined') {
+    if (navigator.languages?.length) {
+      return navigator.languages[0];
+    }
+
+    if (navigator.language) {
+      return navigator.language;
+    }
+  }
+};
```
KosmosKey, line 13: I've done some additional checks in here to make sure it works with SSR too. Here you will either get array of languages which means we will pick the first one.

Otherwise if array of languages doesn't exist, but we have a single language (`navigator.language`) then we can choose that ✅

For older browsers like internet explorer this will return `undefined` if they don't support `navigator`.

packages/example/utils/locale.ts
```diff
@@ -0,0 +1,51 @@
...
+    case 'ja-JP':
+      return Locale.JA_JP;
+
+    case 'pt-BR':
+      return Locale.PT_BR;
+
+    case 'zh-CN':
+      return Locale.ZH_CN;
+
+    case 'id-ID':
```
KosmosKey, line 36: Bunch of `switch` cases based off of browsers `locale`. We could go ahead and change our `Locale` enum to match browsers `locale`, but better to use underscores `_` like `en_US` 🙌

packages/rainbowkit/src/components/RainbowKitProvider/I18nContext.tsx
```diff
@@ -0,0 +1,19 @@
...
+}
+
+export const I18nProvider = ({ children, locale }: I18nProviderProps) => {
+  const i18n = useMemo(() => {
+    if (locale) _i18n.locale = locale;
+    return _i18n;
+  }, [locale]);
+
+  return <I18nContext.Provider value={i18n}>{children}</I18nContext.Provider>;
+};
```
KosmosKey, line 19: This will be the `I18n` provider / context we will use through out the app. Even tho we can go for singleton changes React still need to re render the `i18n` value when `locale` prop changes.

When `locale` prop changes we will also change the `i18n` singleton from `locales.ts` to the new language user choose.

packages/rainbowkit/src/components/RainbowKitProvider/RainbowKitProvider.tsx
```diff
@@ -105,56 +109,58 @@ export function RainbowKitProvider({
...
     <RainbowKitChainProvider chains={chains} initialChain={initialChain}>
-      <CoolModeContext.Provider value={coolMode}>
-        <ModalSizeContext.Provider
-          value={isSmallScreen ? ModalSizeOptions.COMPACT : modalSize}
-        >
-          <ShowRecentTransactionsContext.Provider
-            value={showRecentTransactions}
+      <I18nProvider locale={locale}>
+        <CoolModeContext.Provider value={coolMode}>
+          <ModalSizeContext.Provider
```
KosmosKey, line 114: Just wrapping up everything in a `I18n` provider

packages/rainbowkit/src/wallets/walletConnectors/enkryptWallet/enkryptWallet.ts
```diff
@@ -59,21 +59,21 @@ export const enkryptWallet = ({
...
-                title: 'Install the Enkrypt Wallet extension',
+                title: 'wallet_connectors.extension.enkrypt.step1.title',
               },
               {
                 description:
-                  'Be sure to back up your wallet using a secure method. Never share your secret phrase with anyone.',
+                  'wallet_connectors.extension.enkrypt.step2.description',
                 step: 'create',
-                title: 'Create or Import a Wallet',
+                title: 'wallet_connectors.extension.enkrypt.step2.title',
```
KosmosKey, line 70: Instead of importin `i18n` and then doing `i18n.t()` we will specify these instructions as key strings. So when `connectorsForWallets` is getting executed we will take those keys and use `i18n` in `connectorsForWallets` 🙏

packages/rainbowkit/src/wallets/connectorsForWallets.ts
```diff
@@ -124,6 +124,38 @@ export const connectorsForWallets = (walletList: WalletList) => {
...
+              steps: walletInstance.extension.instructions.steps.map(
+                (step) => ({
+                  ...step,
+                  title: i18n.t(step.title),
+                  description: i18n.t(step.title),
+                }),
+              ),
+            },
+          };
+        }
```
KosmosKey, line 157: Pretty striaght forward. If `instructions` exists in `qrCode` or in `extension` we will make sure the steps are localized and translated based off of their string keys: https://github.com/rainbow-me/rainbowkit/pull/1534/files#r1341738637

packages/example/pages/_app.tsx
```diff
@@ -246,9 +254,36 @@ function RainbowKitApp({
 
   const [isMounted, setIsMounted] = useState(false);
   useEffect(() => setIsMounted(true), []);
+  // biome-ignore lint/nursery/useExhaustiveDependencies: TODO
+  useEffect(() => {
+    wagmiConfig.setConnectors(connectors);
+  }, [locale]);
```
KosmosKey, line 260: ✨ In the singleton approach with `i18n`, connectors run only once and aren’t updated thereafter. For every page load or locale change, we instruct wagmi to reset the connectors, enabling `connectorsForWallets` to reflect the latest `i18n` singleton class locale 🙌. 

Through `useMemo` from `I18nProvider` will update singleton locale from `i18n` throughout the app's runtime. https://github.com/rainbow-me/rainbowkit/pull/1534/files#r1341737308

KosmosKey: I tried to do this force refresh connector logic in `RainbowKitProvider` by using `useConfig` hook and `setConnector` function from wagmi, but it crashes and doesn't seem to be ready yet.

packages/example/pages/_app.tsx
DanielSinclair: Snooped at other approaches here. I think we can accept the raw string directly like "en-US". How about we expose a union type for that prop similar to `Theme` and export as a type rather than an enum
```ts
export type Locale =
  | 'en-US'
  | 'es-419'
  | 'fr-FR'
  | 'ja-JP'
  | 'pt-BR'
  | 'zh-CN'
  | 'id-ID'
  | 'hi-IN'
  | 'tr-TR'
  | 'ru-RU';
```
Researched a bit more about Next.js, and I think it makes sense for us to compatible with their `en-US` format instead of `en_US` externally. But we should keep using our format internally. I think we'd should have to bring your switch case into the repo which would be good to simplify the example too.

packages/example/utils/locale.ts
DanielSinclair: Reference this above; think we should bring this into the package to simplify Next.js use

packages/example/next.config.js
```diff
@@ -1,18 +1,18 @@
...
+      'es-419',
+      'fr-FR',
+      'ja-JP',
+      'pt-BR',
+      'zh-CN',
+      'id-ID',
+      'hi-IN',
+      'tr-TR',
+      'ru-RU',
     ],
```
KosmosKey, line 16: This will be `locales` that the browser will support

packages/example/next.config.js
```diff
@@ -1,4 +1,19 @@
...
+      'fr-FR',
+      'ja-JP',
+      'pt-BR',
+      'zh-CN',
+      'id-ID',
+      'hi-IN',
+      'tr-TR',
+      'ru-RU',
+    ],
+  },
```
KosmosKey, line 17: These locales will be the locales the the browsers support

packages/example/pages/_app.tsx
```diff
@@ -207,12 +212,29 @@ type OverlayBlur = typeof overlayBlurs[number];
...
+  'pt-BR',
+  'zh-CN',
+  'id-ID',
+  'hi-IN',
+  'tr-TR',
+  'ru-RU',
+] as const;
+
+export type Locale = typeof locales[number];
+
```
KosmosKey, line 230: Making a type here exactly like we do with `ModalSize`

packages/example/pages/_app.tsx
```diff
@@ -245,10 +268,45 @@ function RainbowKitApp({
...
+    }
+
+    return locale;
+  };
+
+  // biome-ignore lint/nursery/useExhaustiveDependencies: TODO
+  useEffect(() => {
+    const browserDetectedLocale = getUserDetectedBrowserLocale();
+    setLocale(browserDetectedLocale as Locale);
+  }, []);
```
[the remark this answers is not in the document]
KosmosKey: For now this is in the `packages/example` 🙏

packages/rainbowkit/src/locales/index.ts
```diff
@@ -0,0 +1,41 @@
...
+  | 'es-419'
+  | 'fr-FR'
+  | 'ja-JP'
+  | 'pt-BR'
+  | 'zh-CN'
+  | 'id-ID'
+  | 'hi-IN'
+  | 'tr-TR'
+  | 'ru-RU';
+
```
KosmosKey, line 24: These will be `Locale` types that browsers support 🙏

packages/rainbowkit/src/locales/index.ts
```diff
@@ -0,0 +1,41 @@
...
+  ['es-419']: es_419,
+  ['fr-FR']: fr_FR,
+  ['hi_IN']: hi_IN,
+  ['id-ID']: id_ID,
+  ['ja-JP']: ja_JP,
+  ['pt-BR']: pt_BR,
+  ['ru-RU']: ru_RU,
+  ['tr-TR']: tr_TR,
+  ['zh-CN']: zh_CN,
+});
```
KosmosKey, line 36: Based off of the browser `Locale` type we will let `i18n` know which language it should translate to.

packages/example/pages/_app.tsx
```diff
@@ -294,6 +356,30 @@ function RainbowKitApp({
                 <h3>RainbowKitProvider props</h3>
                 <table cellSpacing={12}>
                   <tbody>
+                    <tr>
+                      <td>
+                        <label
+                          htmlFor="authEnabled"
```
DanielSinclair, line 362: We can drop this tag; was a styling override IIRC

packages/rainbowkit/src/components/ConnectOptions/ConnectDetails.tsx
```diff
@@ -196,19 +200,21 @@ export function ConnectDetail({
     href?: string;
   } | null = showWalletConnectModal
     ? {
-        description: `Need the ${
-          compactModeEnabled ? '' : 'official'
-        } WalletConnect modal?`,
-        label: 'OPEN',
+        description: `${i18n.t('intro.wallet.get_started.need')} ${
+          compactModeEnabled ? '' : i18n.t('intro.wallet.get_started.official')
+        } ${i18n.t('intro.wallet.get_started.wallet_connect')}`,
```
DanielSinclair, line 205: We need to bring these variables into the string for RTL languages/parsing; you can use multiple variables like this; just need to pass `modifier` to the i18n.t function
```
Need the %{modifier}WalletConnect modal?
```
Logic:
```ts
compactModeEnabled ? 'official ' : '' for proper spacing
```

packages/rainbowkit/src/components/ConnectOptions/ConnectDetails.tsx
```diff
@@ -196,19 +200,21 @@ export function ConnectDetail({
...
         onClick: () => {
           onClose();
           showWalletConnectModal();
         },
       }
     : hasQrCode
     ? {
-        description: `Don\u2019t have ${name}?`,
-        label: 'GET',
+        description: `${i18n.t('intro.wallet.get_started.not_found', {
```
DanielSinclair, line 214: We should bring over `\u2019` to the string; I believe this was done to prevent a style system issue. There are a few other instances of this; shouldn't have a problem translating

packages/rainbowkit/src/components/ConnectOptions/ConnectDetails.tsx
```diff
@@ -606,8 +620,10 @@ export function DownloadOptionsDetail({
       >
         {extensionDownloadUrl && (
           <DownloadOptionsBox
-            actionLabel={`Add to ${browser}`}
-            description="Access your wallet right from your favorite web browser."
+            actionLabel={`${i18n.t(
+              'intro.wallet.get_started.add_to',
+            )} ${browser}`}
```
DanielSinclair, line 625: String replacement should look like this:
```ts
"Add to %{browser}"
i18n.t("intro.wallet.get_started.add_to", { browser });
```
We'll want to correct any instances of this

packages/rainbowkit/src/components/ConnectOptions/ConnectDetails.tsx
```diff
@@ -666,9 +688,9 @@ export function DownloadDetail({
       height="full"
       width="full"
     >
-      <Box style={{ maxWidth: 220, textAlign: 'center' }}>
+      <Box style={{ textAlign: 'center' }}>
```
DanielSinclair, line 691: We should keep the maxWidth here; any reason we dropped it? New-line overflow is usually the better option

packages/rainbowkit/src/components/ConnectOptions/DesktopOptions.tsx
```diff
@@ -354,7 +369,11 @@ export function DesktopOptions({ onClose }: { onClose: () => void }) {
                           size="14"
                           weight="bold"
                         >
-                          {groupName}
+                          {supportedI18nGroupNames.includes(groupName)
+                            ? i18n.t(
+                                `intro.group_name.${groupName.toLowerCase()}`,
+                              )
+                            : groupName}
```
DanielSinclair, line 376: Nice 👌

packages/rainbowkit/src/locales/index.ts
DanielSinclair, line 36: `'en-US'` should work here

packages/rainbowkit/src/wallets/connectorsForWallets.ts
DanielSinclair: We actually want to use the `i18n` object downstream in the React tree , rather than in `connectorsForWallets`; that way we don't need a refresh or to prepare connectors more than once. Have a look this implementation; we would just use the key from the connector here:
https://github.com/rainbow-me/rainbowkit/blob/252f02e8c024e2f92734653e8ec5fd2180434b07/packages/rainbowkit/src/components/ConnectOptions/ConnectDetails.tsx#L749

packages/rainbowkit/src/wallets/walletConnectors/argentWallet/argentWallet.ts
```diff
@@ -67,22 +67,19 @@ export const argentWallet = ({
           learnMoreUrl: 'https://argent.xyz/learn/what-is-a-crypto-wallet/',
           steps: [
             {
-              description:
-                'Put Argent on your home screen for faster access to your wallet.',
+              description: 'wallet_connectors.qr_code.argent.step1.description',
               step: 'install',
-              title: 'Open the Argent app',
+              title: 'wallet_connectors.qr_code.argent.step1.title',
```
DanielSinclair, line 72: For maintenance simplicity, we should flip around the keys here to look something like
`wallet_connectors.argent.qr_code.step1.title`

packages/rainbowkit/src/locales/en_US.json
```diff
@@ -0,0 +1,640 @@
...
+      "label": "INSTALL"
+    },
+    "group_name": {
+      "recommended": "Recommended",
+      "other": "Other",
+      "popular": "Popular",
+      "more": "More",
+      "others": "Others"
+    },
+    "wallet": {
```
DanielSinclair, line 25: Let's have a more clear key hierarchy for maintenance. The `intro` key should only cover this UI element. The "Get a wallet" flow should be a separate top-level key. Same for "Connect a Wallet" sidebar, and any drill-down flow. WalletConnectors will have their own top-level key, but the UI flows should otherwise be reflected like other UI flows.
Screenshot 2023-10-02 at 4 30 59 PM

DanielSinclair (changes requested):

packages/rainbowkit/src/components/ConnectOptions/ConnectDetails.tsx
KosmosKey: Ah yeah.

On some languages the break of the text might look bad

image
image

packages/rainbowkit/src/locales/index.ts
KosmosKey: Yeah its already included here ?

image

packages/rainbowkit/src/components/ConnectOptions/ConnectDetails.tsx
KosmosKey: Ah yeah missed that one 👍

packages/rainbowkit/src/components/ConnectOptions/ConnectDetails.tsx
```diff
@@ -1,45 +1,46 @@
...
     case BrowserType.Opera:
-      return (await import('../Icons/Opera.svg')).default;
+      return (await import("../Icons/Opera.svg")).default;
     case BrowserType.Safari:
-      return (await import('../Icons/Safari.svg')).default;
+      return (await import("../Icons/Safari.svg")).default;
     default:
-      return (await import('../Icons/Browser.svg')).default;
+      return (await import("../Icons/Browser.svg")).default;
   }
```
DanielSinclair, line 44: @KosmosKey It seems like you might have a local linter problem; try disabling any eslint plugins and enable this: https://marketplace.visualstudio.com/items?itemName=biomejs.biome

KosmosKey: Will fix that 🙏

packages/rainbowkit/src/components/ConnectModal/ConnectModalIntro.tsx
```diff
@@ -18,92 +19,91 @@ export function ConnectModalIntro({
...
+    <Box
+      alignItems="center"
+      color="accentColor"
+      display="flex"
+      flexDirection="column"
+      justifyContent="space-around"
+      flex="1"
+      overflowY="auto"
+    >
+      <Box marginBottom="10">
```
KosmosKey, line 34: Ignore a lot of these changes. We used to have some `height` and `min-height` tricks to make rainbowkit look suitable but right now it doesn't support responsive design. So i use flexbox and overflow to make this work 🙏

Example:

image

packages/rainbowkit/src/components/ConnectButton/ConnectButton.tsx
```diff
@@ -39,6 +40,14 @@ export function ConnectButton({
...
   const chains = useRainbowKitChains();
   const connectionStatus = useConnectionStatus();
+  const [isClient, setIsClient] = useState(false);
+
+  const i18n = useContext(I18nContext);
+
+  // Quick fix to avoid hydration errors with next-js on brave browser
+  useEffect(() => {
+    setIsClient(true);
+  }, []);
```
KosmosKey, line 50: Added this avoid rehydration error from nextjs on other brave browser when not providing `locale` prop. Just simple hack.

packages/rainbowkit/src/components/RainbowKitProvider/I18nContext.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+  const i18n = useMemo(() => {
+    if (locale) {
+      _i18n.locale = locale;
+    } else if (!locale && detectedBrowserLocale) {
+      _i18n.locale = detectedBrowserLocale;
+    }
+
+    return _i18n;
+  }, [locale, detectedBrowserLocale]);
```
KosmosKey, line 34: If `locale` prop is provided -> use the `locale` prop
If `locale` is not provided and `detectedBrowserLocale` exist -> use `detectedBrowserLocale` as a default locale for the user

Otherwise return normal `_i18n` class with `en-US` locale 🙏

packages/rainbowkit/src/components/ConnectButton/ConnectButton.tsx
DanielSinclair: @KosmosKey What issue does this resolve? Is this a SSR issue?

packages/rainbowkit/src/components/ConnectModal/ConnectModalIntro.tsx
DanielSinclair: Nice 👌

packages/rainbowkit/src/components/ConnectButton/ConnectButton.tsx
KosmosKey: Yeah it's an SSR issue. It will look like this:

KosmosKey: @DanielSinclair I wanted to remove this early today, but dev people will have hard time on this an open ticket which might lead to more github issues.

I can explain what is happening so we might keep it as is:
- When doing browser detection in SSR next.js has `Connect Wallet` while running build
- On CSR the language detection updates the locale to be something like `वॉलेट को कनेक्ट करें`

Next.js is then getting confused but this won't affect prod but while devs are working it might impact them.

packages/rainbowkit/src/utils/locale.ts
```diff
@@ -26,7 +26,50 @@ export const locales = [
...
+  }
+
+  if (isIncluded(['tr', 'tr-TR'], locale)) {
+    return 'tr-TR';
+  }
+
+  if (isIncluded(['ru', 'ru-RU'], locale)) {
+    return 'ru-RU';
+  }
 };
```
DanielSinclair, line 75: A bit of a cleaner way to implement this:
```ts
export const standardizeBrowserLocale = (
  locale: string,
): Locale | undefined => ({
  'ru-RU': 'ru-RU',
  'ru': 'ru-RU', 
})[locale] as Locale;
```

KosmosKey: Nice! Yeah object structuring s the best way to go instead of repeating the same function over and over again 👍

https://github.com/rainbow-me/rainbowkit/pull/1534/commits/241a70739ce3a07223e327382b8ab7ff686b4420#diff-97daf67abd2e298c4f7264deb2b51e364f392924ad5f82a793ff38dfb75544daR33-R53

packages/example/pages/_app.tsx
```diff
@@ -245,10 +263,14 @@ function RainbowKitApp({
       : currentTheme.accentColors[selectedAccentColor];
 
   const [isMounted, setIsMounted] = useState(false);
+
   useEffect(() => setIsMounted(true), []);
 
   const appContextProps: AppContextProps = { authEnabled };
 
+  const locales = router.locales!;
+  const locale = (router.locale || router.defaultLocale) as Locale;
```
DanielSinclair, line 272: Is `locales` here being used? Conflicts with the drop-down locales list variable

packages/example/pages/_app.tsx
```diff
@@ -474,7 +474,9 @@ function RainbowKitApp({
                       <td>
                         <select
                           onChange={(e) => {
-                            setLocale(e.target.value as Locale);
+                            router.push('/', undefined, {
+                              locale: e.target.value,
+                            });
```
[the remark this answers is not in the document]
DanielSinclair: @KosmosKey We should actually keep both approaches, I think. We essentially need to test what happens when you change the state on re-render for Create React App as well as URL subpaths for Next.js

packages/example/pages/_app.tsx
```diff
@@ -207,12 +208,29 @@ type OverlayBlur = typeof overlayBlurs[number];
...
+  'pt-BR',
+  'zh-CN',
+  'id-ID',
+  'hi-IN',
+  'tr-TR',
+  'ru-RU',
+] as const;
+
+export type Locale = typeof locales[number];
+
```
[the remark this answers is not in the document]
KosmosKey: No longer since we support rainbowkit `Locale`

packages/example/pages/_app.tsx
```diff
@@ -227,6 +231,11 @@ function RainbowKitApp({
   const [showDisclaimer, setShowDisclaimer] = useState(false);
   const [customAvatar, setCustomAvatar] = useState(false);
 
+  const routerLocale = router.locale as Locale;
+
+  // Set `locale` as default from next.js and let dropdown set new `locale`
+  const [locale, setLocale] = useState<Locale>(routerLocale);
+
```
DanielSinclair, line 238: This works super well. Nice 👌

packages/rainbowkit/src/components/ConnectButton/ConnectButton.test.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+  it("Displays in English for 'en-US'", async () => {
+    const text = await renderTextButton('en-US');
+    expect(text).toBe('Connect Wallet');
+  });
+
+  it("Displays in Spanish for 'es-419'", async () => {
+    const text = await renderTextButton('es-419');
+    expect(text).toBe('Conectar la billetera');
+  });
```
KosmosKey, line 38: Tests for `<ConnectButton />` ✨

packages/rainbowkit/src/components/ConnectModal/ConnectModal.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+  it('Defaults to English without a `locale` prop', async () => {
+    const label = await renderHeaderLabelModal();
+    expect(label).toBe('Connect a Wallet');
+  });
+
+  it("Displays in English for 'en-US'", async () => {
+    const label = await renderHeaderLabelModal('en-US');
+    expect(label).toBe('Connect a Wallet');
+  });
```
KosmosKey, line 36: ✨ Tests for `<ConnectModal />`.

I was also thinking of doing like this:
```
i18n.locale = "ru-RU"
const label = await renderHeaderLabelModal('en-US');
expect(label).toBe(i18n.t("keyword"));
```

But will require more code ofc, tried to keep it clean in this case 🤷‍♂️

DanielSinclair:
Spotted an issue when hovering over the GET button here. Reverted styling changes because they were a bit over my head, but we'll reintroduce and fix a few more styling in a patch tomorrow/next week
Screenshot 2023-10-06 at 1 16 17 AM

DanielSinclair (approved):
Shipping 🚀

PR #1553, opened 2023-10-11
fix: sign out if wallet is switched

0xmkh:
Whenever user does a wallet switch he needs to be logged out immediately. This is not what's happening as of now.

### DIFF
--- a/.changeset/polite-maps-decide.md
+++ b/.changeset/polite-maps-decide.md
@@ -2,4 +2,4 @@
 "@rainbow-me/rainbowkit": patch
 ---
 
-Implemented logout on wallet switch
+Implemented logout on wallet switch in `<RainbowKitAuthenticationProvider />`
--- a/.changeset/polite-maps-decide.md
+++ b/.changeset/polite-maps-decide.md
@@ -2,4 +2,4 @@
 "@rainbow-me/rainbowkit": patch
 ---
 
-Implemented logout on wallet switch in `<RainbowKitAuthenticationProvider />`
+Fixed an issue where a user would not automatically be logged out after switching their wallet in MetaMask or other browser wallets
--- a/.changeset/polite-maps-decide.md
+++ b/.changeset/polite-maps-decide.md
@@ -2,4 +2,4 @@
 "@rainbow-me/rainbowkit": patch
 ---
 
-Fixed an issue where a user would not automatically be logged out after switching their wallet in MetaMask or other browser wallets
+Fixed an issue where a user would not get automatically logged out from the Authentication API after switching their wallet in MetaMask or other browser wallets. Users must now sign a new SIWE message after switching wallets.
--- a/.changeset/polite-maps-decide.md
+++ b/.changeset/polite-maps-decide.md
@@ -1,0 +1,5 @@
+---
+"@rainbow-me/rainbowkit": patch
+---
+
+Implemented logout on wallet switch
--- a/packages/rainbowkit/src/components/RainbowKitProvider/AuthenticationContext.tsx
+++ b/packages/rainbowkit/src/components/RainbowKitProvider/AuthenticationContext.tsx
@@ -6,12 +6,7 @@ import React, {
   useMemo,
   useRef,
 } from 'react';
-import { useAccount } from 'wagmi';
-
-interface ConnectorData {
-  account: string;
-  chain: number;
-}
+import { ConnectorData, useAccount } from 'wagmi';
 
 export type AuthenticationStatus =
   | 'loading'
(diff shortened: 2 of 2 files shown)


packages/rainbowkit/src/components/RainbowKitProvider/AuthenticationContext.tsx
```diff
@@ -55,7 +60,7 @@ export function RainbowKitAuthenticationProvider<Message = unknown>({
 }: RainbowKitAuthenticationProviderProps<Message>) {
   // When the wallet is disconnected, we want to tell the auth
   // adapter that the user session is no longer active.
-  useAccount({
+  const { connector } = useAccount({
     onDisconnect: () => {
       adapter.signOut();
     },
```
DanielSinclair, line 66: @KosmosKey Wagmi also has an optional `onConnect` callback here. That might simplify this if it is fired when the address changes

DanielSinclair, line 66: @KosmosKey Wagmi also has an optional `onConnect` callback here. That might simplify this if it is fired when the address changes

DanielSinclair: https://wagmi.sh/react/hooks/useContractEvent#usage

KosmosKey:
Here are some thoughts on github issues from wagmi:
- [example 1](https://github.com/wagmi-dev/wagmi/discussions/1806#discussioncomment-7113585)
- [example 2](https://github.com/wagmi-dev/wagmi/discussions/1806#discussioncomment-5422081)

packages/rainbowkit/src/components/RainbowKitProvider/AuthenticationContext.tsx
```diff
@@ -77,6 +82,24 @@ export function RainbowKitAuthenticationProvider<Message = unknown>({
...
+    // Only if account is changed then signOut
+    if (account) adapter.signOut();
+  };
+
+  // biome-ignore lint/nursery/useExhaustiveDependencies: <explanation>
+  useEffect(() => {
+    if (connector && status === 'authenticated') {
+      // Attach the event listener when status is 'authenticated'
+      connector.on('change', handleChangedAccount);
+
```
KosmosKey, line 95: Also one thing here. We don't want to listen to `"change"` event unless is logged in with SIWE. If we listen to the event right after user connects his wallet then we can't log him out since he is not authenticated yet. 

Let me know if this ok approach 👍

.changeset/polite-maps-decide.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rainbow-me/rainbowkit": patch
+---
+
+Implemented logout on wallet switch in `<RainbowKitAuthenticationProvider />`
```
DanielSinclair, line 5: Might say something more specific like "Fixed an issue where a user would not automatically be logged out after switching their wallet in MetaMask or other browser wallets"

packages/rainbowkit/src/components/RainbowKitProvider/AuthenticationContext.tsx
```diff
@@ -77,6 +82,24 @@ export function RainbowKitAuthenticationProvider<Message = unknown>({
     }
   }, [status, adapter, isDisconnected]);
 
+  const handleChangedAccount = ({ account }: ConnectorData) => {
+    // Only if account is changed then signOut
+    if (account) adapter.signOut();
+  };
+
+  // biome-ignore lint/nursery/useExhaustiveDependencies: <explanation>
+  useEffect(() => {
```
DanielSinclair, line 91: Would add more brevity to comments here to describe exactly why listeners in `useEffect` are more reliable than `onConnect`

.changeset/polite-maps-decide.md
DanielSinclair, line 5: Might say something more specific like "Fixed an issue where a user would not automatically be logged out after switching their wallet in MetaMask or other browser wallets"

packages/rainbowkit/src/components/RainbowKitProvider/AuthenticationContext.tsx
```diff
@@ -8,6 +8,11 @@ import React, {
 } from 'react';
 import { useAccount } from 'wagmi';
 
+interface ConnectorData {
+  account: string;
+  chain: number;
+}
+
```
DanielSinclair, line 15: Can we infer this type import from `@wagmi/core` directly?

DanielSinclair:
Functionality is working well 👍

packages/rainbowkit/src/components/RainbowKitProvider/AuthenticationContext.tsx
KosmosKey: Yup adding it 🙏

packages/rainbowkit/src/components/RainbowKitProvider/AuthenticationContext.tsx
KosmosKey: Added their `ConnectorData` type. Here is the type for more info: [wagmi type](https://github.com/wagmi-dev/wagmi/blob/38306606d2fd72a4c6918323bf86a1afda348638/packages/connectors/src/base.ts#L14)

DanielSinclair (approved):

PR #1556, opened 2023-10-11
feat: add holesky testnet

0xmkh:


DanielSinclair:
Works for me 👍 
Screenshot 2023-10-11 at 7 15 29 PM

DanielSinclair:
Switched the chain list ordering a bit so that it's easier to maintain in the long-run, and also split up the changeset so that we update the `package/example` changelog and `package/rainbowkit` changelog separately

DanielSinclair (approved):

PR #1559, opened 2023-10-13
fix: `alchemyProvider` params

1337-ishaan:
Fixed, 

`alchemyProvider({ alchemyId: process.env.ALCHEMY_ID }) --> alchemyProvider({ apiKey: process.env.ALCHEMY_ID }),`


DanielSinclair:
Hey! These are changes to the historical changelog and migration guide, so they shouldn't be necessary. Let me know if I missed anything!

1337-ishaan:
image

Hey @DanielSinclair, I think you need to look at this as the parameter passed here is `alchemyId`, but originally it is `apiKey`

PR #1561, opened 2023-10-15
Add XDEFI Wallet provider

zdeadex:


DanielSinclair:
@zdeadex Unfortunately we won't be expanding the default wallet list, but RainbowKit will support EIP-6963 very soon, so we recommend that you instead adopt this in XDEFI for the best user experience.

zdeadex:
> @zdeadex Unfortunately we won't be expanding the default wallet list, but RainbowKit will support EIP-6963 very soon, so we recommend that you instead adopt this in XDEFI for the best user experience.

We support it already, thank's for the feedback!

PR #1569, opened 2023-10-17
feat: rainbow button

0xmkh:


packages/button/package.json
```diff
@@ -0,0 +1,41 @@
...
+    "build:watch": "node build.js --watch",
+    "dev": "pnpm build:watch & pnpm typegen:watch",
+    "prepare": "pnpm build",
+    "prebuild": "pnpm typegen",
+    "typecheck": "pnpm tsc --noEmit",
+    "typegen": "tsc --emitDeclarationOnly || true",
+    "typegen:watch": "tsc --emitDeclarationOnly --watch"
+  },
+  "peerDependencies": {
+    "@rainbow-me/rainbowkit": ">=1.1.1",
```
DanielSinclair, line 29: We'll have to keep this in mind and first deploy the RainbowKit changes and then bump up this peer dependency

packages/button/package.json
```diff
@@ -0,0 +1,41 @@
+{
+  "name": "@rainbow-me/button",
+  "description": "Rainbow connector button",
+  "version": "0.0.1",
```
DanielSinclair, line 4: We can bump this to `0.1.0`

packages/example/pages/index.tsx
```diff
@@ -250,6 +251,70 @@ const Example = ({ authEnabled }: AppContextProps) => {
...
+            }}
+          >
+            <RainbowButton />
+          </div>
+        </div>
+      </div>
+
+      <div>
+        <h3 style={{ fontFamily: 'sans-serif' }}>Rainbow button (custom)</h3>
+        <div
```
DanielSinclair, line 276: We could bump this up to their "Custom buttons" section. We can probably skip the styling too just to reduce testable surface area

examples/with-next-rainbow-button/pages/_app.tsx
```diff
@@ -0,0 +1,34 @@
...
+  [mainnet],
+  [publicProvider()]
+);
+
+const projectId =
+  process.env.NEXT_PUBLIC_WALLETCONNECT_PROJECT_ID ?? "YOUR_PROJECT_ID";
+
+const wagmiClient = createConfig({
+  autoConnect: true,
+  connectors: [new RainbowConnector({ chains, projectId })],
```
DanielSinclair, line 19: This dev UX isn't too bad; I think it'll work

KosmosKey: Yeah right. I was looking at handling Context API, but that's way complicated. I think most devs are used to importing connectors in their project so they can pretty much do it easily like that as it is.

packages/button/src/components/RainbowButton.tsx
```diff
@@ -0,0 +1,14 @@
+import { RainbowKitProvider } from '@rainbow-me/rainbowkit';
+import type { RainbowKitProviderProps } from '@rainbow-me/rainbowkit/dist/components/RainbowKitProvider/RainbowKitProvider';
+import React from 'react';
+
+export function RainbowButtonProvider({
+  children,
+  ...options
+}: Omit<RainbowKitProviderProps, 'chains'>) {
+  return (
+    <RainbowKitProvider chains={[]} {...options}>
```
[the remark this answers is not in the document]
KosmosKey: Yeah will add comment to that.

packages/rainbowkit/src/components/RainbowButton/RainbowButton.tsx
```diff
@@ -0,0 +1,91 @@
+import React, { useState } from 'react';
+import { touchableStyles } from '../../css/touchableStyles';
+import { useWalletConnectors } from '../../wallets/useWalletConnectors';
+import { AsyncImage } from '../AsyncImage/AsyncImage';
+import { Box } from '../Box/Box';
+import * as styles from './RainbowButton.css';
+import { WalletButtonRenderer } from './WalletButtonRenderer';
+
+export const RainbowButton = () => {
+  const [isMouseOver, setIsMouseOver] = useState<boolean>(false);
```
DanielSinclair, line 10: Let's keep this private API as `WalletButton` so that the user could use it for MetaMask and other connectors too

packages/button/src/index.ts
```diff
@@ -0,0 +1,3 @@
+export { RainbowButtonProvider } from './components/RainbowButton';
+export { RainbowConnector } from './connectors/rainbow';
+export { RainbowButton } from '@rainbow-me/rainbowkit';
```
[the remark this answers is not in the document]
KosmosKey: Oh yeah forgot about it. I'll expose the `RainbowButton` as a component rather than importing it from rainbowkit 🙏

packages/button/src/connectors/rainbow.ts
```diff
@@ -0,0 +1,20 @@
...
+}
+
+function RainbowConnector({ chains, projectId }: RainbowWalletOptions) {
+  return connectorsForWallets([
+    {
+      groupName: 'Popular',
+      wallets: [rainbowWallet({ chains, projectId })],
+    },
+  ])()[0];
+}
```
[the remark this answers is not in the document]
KosmosKey: Yeah that's a good point. Essentially even if we use `getDefaultWallets` it will kind of still rely on `rainbowWallet` rainbow connector and we still have to play around with indexes like `[0]` or `.filter()`. 

What about in the next release creating a function that prepares only one connector for us so it's easier and maintainable ?

packages/button/src/components/RainbowButton.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+export function RainbowButtonProvider({
+  children,
+  ...options
+}: Omit<RainbowKitProviderProps, 'chains'>) {
+  // NOTE: The chains array is intentionally left empty.
+  // This is due to the current incompatibility between the 'rainbowkit' chain modal and the 'rainbow' button library.
+  // As a workaround, users can utilize 'wagmi' to switch chains.
+  return (
+    <RainbowKitProvider chains={[]} {...options}>
```
KosmosKey, line 13: Basically we don't need `chains` array since we are not using the rainbowkit chain modal UI here. Ok to leave it empty for now. Since our focus with rainbow button is to only do `connect`.

packages/rainbowkit/src/components/WalletButton/WalletButton.tsx
```diff
@@ -0,0 +1,107 @@
...
+                    >
+                      <Box>{connector?.name}</Box>
+                    </Box>
+                  </Box>
+                </Box>
+              )}
+            </Box>
+
+            {error ? <Box marginTop="8">{error}</Box> : null}
+          </Box>
```
KosmosKey, line 100: I also want to do error handling for the rainbow button. For now i just display a text, but maybe we can add a `x` close button to remove the error ?

packages/rainbowkit/src/components/WalletButton/WalletButtonRenderer.tsx
```diff
@@ -0,0 +1,93 @@
...
+}
+
+export function WalletButtonRenderer({ children }: WalletButtonRendererProps) {
+  const mounted = useIsMounted();
+  const { openConnectModal } = useConnectModal();
+  const { connectModalOpen } = useModalState();
+  const { connector, setConnector } = useContext(RainbowButtonContext);
+  const [firstConnector] = useWalletConnectors()
+    .filter((wallet) => wallet.ready)
+    .sort((a, b) => a.groupIndex - b.groupIndex);
```
KosmosKey, line 34: @DanielSinclair, for `WalletButton` API connector sharing logic, when given connectors like `connectors: [metamaskWallet()]`, we'll always use the first one. This is the same for connectors: `[new RainbowConnector()]`. We can't pass only one connector to the API. 

If developers don't want the first choice, they might need a specific `connector` prop which they can pass in connector themselves. But generally, this works fine. Let me know what you think.

packages/button/package.json
```diff
@@ -0,0 +1,41 @@
...
+  "files": [
+    "dist"
+  ],
+  "keywords": [],
+  "author": "Rainbow",
+  "license": "MIT",
+  "type": "module",
+  "exports": {
+    ".": "./dist/index.js",
+    "./styles.css": "./node_modules/@rainbow-me/rainbowkit/styles.css"
```
DanielSinclair, line 14: Because `rainbowkit` is a peer dependency, this might be an issue. I think we should bundle this with `/dist` in the `built.js` script; that should always work at build-time because rainbowkit is a dev dependency

packages/button/package.json
```diff
@@ -0,0 +1,41 @@
...
+    "build:watch": "node build.js --watch",
+    "dev": "pnpm build:watch & pnpm typegen:watch",
+    "prepare": "pnpm build",
+    "prebuild": "pnpm typegen",
+    "typecheck": "pnpm tsc --noEmit",
+    "typegen": "tsc --emitDeclarationOnly || true",
+    "typegen:watch": "tsc --emitDeclarationOnly --watch"
+  },
+  "peerDependencies": {
+    "@rainbow-me/rainbowkit": "1.0.x || 1.1.x",
```
DanielSinclair, line 29: Let's make this `1.2.x`

packages/rainbowkit/src/components/ConnectOptions/DesktopOptions.tsx
```diff
@@ -16,6 +16,7 @@ import { InfoButton } from '../InfoButton/InfoButton';
 import { ModalSelection } from '../ModalSelection/ModalSelection';
 import { AppContext } from '../RainbowKitProvider/AppContext';
 import { I18nContext } from '../RainbowKitProvider/I18nContext';
+
```
DanielSinclair, line 19: Spooky new line

packages/rainbowkit/src/components/ConnectOptions/ConnectOptions.tsx
```diff
@@ -1,10 +1,16 @@
...
-  return isMobile() ? (
+  const { connector } = useContext(RainbowButtonContext);
+
+  // If you do inspect element on mobile then switch to
+  // desktop view the rainbowkit connect modal will pop up.
+  // In this case we don't want that if we are using a custom connector
+  return isMobile() || connector ? (
     <MobileOptions onClose={onClose} />
   ) : (
     <DesktopOptions onClose={onClose} />
```
DanielSinclair, line 16: We would do a switch here to `MobileLinking`

packages/rainbowkit/src/components/ConnectOptions/MobileOptions.tsx
DanielSinclair, line 1: I think we should break this flow out into an independent component because it will add complexity to this already-heavy component. Maybe something like `MobileLinking/MobileLinking.tsx`

packages/rainbowkit/src/components/ConnectOptions/MobileOptions.tsx
```diff
@@ -150,6 +212,7 @@ function WalletButton({
 enum MobileWalletStep {
   Connect = 'CONNECT',
   Get = 'GET',
+  WalletConnect = 'WALLETCONNECT',
 }
```
DanielSinclair, line 216: We shouldn't need to modify this control tree if we route to this component at the top-level in `ConnectOptions`

packages/rainbowkit/src/components/WalletButton/WalletButtonRenderer.tsx
DanielSinclair: We'll have to work through this one. If the user adopts the `WalletButton` component, they'll need to pass through a connector id like `<WalletButton wallet='metamask'/>` and we'll need to select the connector through that path. The user could use multiple different WalletButtons simultaneously

packages/example/pages/index.tsx
```diff
@@ -159,95 +160,132 @@ const Example = ({ authEnabled }: AppContextProps) => {
...
           }}
-        </ConnectButton.Custom>
+        >
+          <div
+            style={{
+              display: 'flex',
+              flexDirection: 'column',
+            }}
+          >
+            <RainbowButton />
```
DanielSinclair, line 286: We should add `<WalletButton wallet='metaMask'/>` here to test the stateful behavior of two buttons alongside each other (which some devs might do for their wallet list)

packages/example/pages/index.tsx
```diff
@@ -159,95 +160,132 @@ const Example = ({ authEnabled }: AppContextProps) => {
...
+                  disabled={!ready}
+                  onClick={connect}
+                  style={{ marginLeft: '18px' }}
+                >
+                  Connect Rainbow
+                </button>
+              );
+            }}
+          </RainbowButton.Custom>
+        </div>
```
DanielSinclair, line 268: We should add a custom button for the metamask wallet button here too

packages/rainbowkit/src/components/WalletButton/WalletButtonRenderer.tsx
DanielSinclair: We also can't control the order of the connectors list, as the dApp will use other connectors for their custom injected/walletconnector behavior or other wallet buttons

KosmosKey: Oh yeah i was thinking the same. Passing some sort of prop to the component. I thought we were initally going to pass what's in the `connectors` from wagmi initially.

packages/button/package.json
KosmosKey: 👍 💡

packages/rainbowkit/src/components/ConnectOptions/ConnectOptions.tsx
KosmosKey: Aha makes sense. Can you give more context on that ? I believe we can create a component called `MobileLinking` that does this type of logic ? Is that the case ?

packages/rainbowkit/src/components/ConnectOptions/MobileOptions.tsx
KosmosKey: Ah i see now. Yeah let me fix that. Here is what i will do:

- Create `MobileLinking.tsx` component instead of putting all logic into `MobileOptions.tsx` component
- Pass in a `customConnector` prop to `MobileOptions.tsx` component so that we can have prop inital check instead of importing all stuff into `MobileOptions.tsx`

Finally will update the `ConnectOptions.tsx` to use `MobileLinking.tsx` if there are any custom connector on our end.

packages/rainbowkit/src/components/ConnectOptions/MobileOptions.tsx
KosmosKey: Good point 🙏

packages/example/pages/index.tsx
KosmosKey: Perfect feedback ⚡️. Looking forward for this feature where you can pass any provider ✨

packages/example/pages/index.tsx
KosmosKey: Great idea. Yeah that would be awesome to include custom connectors 🎶 🙌

packages/button/build.js
```diff
@@ -0,0 +1,60 @@
...
+  outdir: 'dist/rainbowkit/styles',
+  watch: isWatching
+    ? {
+        onRebuild(error, result) {
+          if (error) console.error('wallets build failed:', error);
+          else console.log('wallets build succeeded:', result);
+        },
+      }
+    : undefined,
+});
```
KosmosKey, line 52: Not the cleanest way, but this will ensure that we push styling into `/dist` folder so we don't rely on `node_modules` here

packages/button/src/components/RainbowButton.tsx
```diff
@@ -0,0 +1,23 @@
...
+      {children}
+    </RainbowKitProvider>
+  );
+}
+
+export const RainbowButton = ({ wallet }: { wallet?: string }) => {
+  return <WalletButton wallet={wallet} />;
+};
```
KosmosKey, line 21: We can now pass in props like `<RainbowButton wallet="metamask" />`

packages/example/pages/index.tsx
```diff
@@ -159,95 +160,136 @@ const Example = ({ authEnabled }: AppContextProps) => {
...
+                >
+                  Connect Rainbow
+                </button>
+              );
+            }}
+          </RainbowButton.Custom>
+
+          <div style={{ marginLeft: '16px' }}>
+            <RainbowButton wallet="metamask" />
+          </div>
```
KosmosKey, line 271: For now i just put it on the `custom buttons` section

packages/rainbowkit/src/components/ConnectOptions/MobileOptions.tsx
```diff
@@ -40,66 +78,79 @@ function WalletButton({
...
+
+  // biome-ignore lint/nursery/useExhaustiveDependencies: <explanation>
+  useEffect(() => {
+    // When using `reactStrictMode: true` in development mode the useEffect hook
+    // will fire twice. We avoid this by using `useRef` logic here. Works for now.
+    if (isCustomConnector && !initialized.current) {
+      onConnect();
+      initialized.current = true;
+    }
+  }, [isCustomConnector]);
```
KosmosKey, line 144: Since `WalletButton` is just a component it can be good to reuse it on our `MobileLinking` component so we make sure we keep the same functionality as it currently is in rainbowkit

packages/rainbowkit/src/components/ConnectOptions/MobileOptions.tsx
```diff
@@ -122,26 +181,28 @@ function WalletButton({
...
+              as="h2"
+              color={wallet.ready ? 'modalText' : 'modalTextSecondary'}
+              size="13"
+              weight="medium"
+            >
+              {/* Fix button text clipping in Safari: https://stackoverflow.com/questions/41100273/overflowing-button-text-is-being-clipped-in-safari */}
+              <Box as="span" position="relative">
+                {shortName ?? name}
+                {!wallet.ready && ' (unsupported)'}
+              </Box>
```
KosmosKey, line 196: No need for this if we use custom connector since we will have a loading spinner instead on mobile.

DanielSinclair:
On the "Get" Desktop flow, I'm seeing a missing back button Screenshot 2023-10-23 at 11 33 27 AM

DanielSinclair:
When Rainbow is injected, the MetaMask WalletButton does not appear Screenshot 2023-10-23 at 11 36 13 AM

DanielSinclair:
When Rainbow is connected, am seeing an issue where the other WalletButtons will also appear as connected; these should be explicit because devs will use multiple connector buttons next to each other, so we want to keep track of the wallet connected. Let's also use a green indicator here 
Screenshot 2023-10-23 at 11 37 08 AM

DanielSinclair:
On mobile after connection, the WalletButtons are still active/interactable IMG_7894

DanielSinclair:
Let's add the dismiss button to this modal so that if a deep link fails, there is a path out. This is a prop within the parent modal component
Screenshot 2023-10-23 at 11 44 24 AM

packages/rainbowkit/src/wallets/latestWalletId.ts
```diff
@@ -0,0 +1,15 @@
+const storageKey = 'rk-latest-id';
+
```
KosmosKey, line 2: I was thinking of renaming this to something else since it's being used mostly in WalletButton API, but wanna keep the naming consistnent by always starting wtih `rk-`

packages/rainbowkit/src/components/MobileLinking/MobileLinking.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+        <Box
+          display="flex"
+          flexDirection="column"
+          alignItems="center"
+          marginTop="16"
+        >
+          <ActionButton label={'Dismiss'} size="medium" onClick={onClose} />
+        </Box>
+      </Box>
```
KosmosKey, line 54: @DanielSinclair here is the dismiss button for mobile 👍

DanielSinclair: We actually have an API for this already with an X in the top right enabled as a prop of the Modal super class. Let's use that to keep the UI feel
IMG_7904

DanielSinclair:
On mobile, there still seems to be an "active state" after the user connects via WalletConnect
Screenshot 2023-10-24 at 2 53 22 AM

DanielSinclair:
Let's use a green icon here instead; we currently use the yellow for pending transactions IIRC
Screenshot 2023-10-24 at 2 47 57 AM

DanielSinclair:
When playing around with this, it was kinda confusing when a connection is rejected, and there was no way to return the Rainbow button to the default state. I think we should deprecate in favor of the only 2 states being unconnected and connected. This is how most wallet buttons I've encountered in dApps work. dApps can still throw a pop-up or something if `isError` is returned from Wagmi's `useConnect` hook.
Screenshot 2023-10-24 at 2 47 11 AM

DanielSinclair:
The Coinbase Wallet connector seems to escape that avoids throwing this modal from the WalletButton entrypoint Screenshot 2023-10-24 at 2 59 22 AM

DanielSinclair:
When stress testing with Rainbow, MetaMask, and Coinbase Wallet injected, I see an issue where the MetaMask button isn't functional and displays the connection error. The other 2 buttons will work and route to the correct provider. Not sure what's causing this 🤔 
Screenshot 2023-10-24 at 3 07 20 AM

KosmosKey:
> On mobile, there still seems to be an "active state" after the user connects via WalletConnect Screenshot 2023-10-24 at 2 53 22 AM

I believe that's supposed to be like that ? If you're on mobile then click on rainbow button and get connected the badge green badge status should say you have connected with rainbow via "walletconnect". Is that the case ?

packages/rainbowkit/src/wallets/walletConnectors/coinbaseWallet/coinbaseWallet.ts
```diff
@@ -8,13 +8,55 @@ export interface CoinbaseWalletOptions {
...
+
+function getCoinbaseWalletInjectedProvider(): Window['ethereum'] {
+  const isCoinbaseWallet = (
+    ethereum: NonNullable<Window['ethereum']> | any,
+  ) => {
+    // Identify if Coinbase Wallet injected provider is present.
+    const coinbaseWallet = !!ethereum.isCoinbaseWallet;
+
+    return coinbaseWallet;
+  };
```
KosmosKey, line 25: Just made sure to add proper check for coinbase wallet since coinbase wallet can be replaced by another injected provider. This way we can see if it's installed properly or not.

packages/rainbowkit/src/locales/en_US.json
```diff
@@ -37,9 +37,11 @@
     "recent": "Recent",
     "status": {
       "opening": "Opening %{wallet}...",
+      "opening_mobile": "Continue in %{wallet}...",
       "not_installed": "%{wallet} is not installed",
       "not_available": "%{wallet} is not available",
-      "confirm": "Confirm connection in the extension"
+      "confirm": "Confirm connection in the extension",
+      "confirm_mobile": "Accept connection request in the wallet"
```
KosmosKey, line 44: Wanna make sure it's nice to include internalization if user are going to use this package with rainbowkit.

packages/rainbowkit/src/components/MobileLinking/MobileLinking.tsx
```diff
@@ -0,0 +1,65 @@
...
+          </Text>
+        </Box>
+
+        <Box maxWidth="310" marginTop="8">
+          <Text textAlign="center" color="modalText" size="16" weight="medium">
+            {i18n.t('connect.status.confirm_mobile', {
+              wallet: connectorName,
+            })}
+          </Text>
+        </Box>
```
KosmosKey, line 59: Also added internalization here for when opining with mobile 🙌

packages/rainbowkit/src/themes/darkTheme.ts
```diff
@@ -58,6 +58,7 @@ export const darkTheme = ({
     profileForeground: 'rgba(224, 232, 255, 0.05)',
     selectedOptionBorder: 'rgba(224, 232, 255, 0.1)',
     standby: '#FFD641',
+    connected: '#00E676',
```
DanielSinclair, line 61: We already have `connectionIndicator` for chains; let's use that color and styling key instead

DanielSinclair:
Moved Coinbase Wallet connector changes to this PR for isolated review: #1582

packages/rainbowkit/src/components/WalletButton/WalletButton.test.tsx
```diff
@@ -0,0 +1,52 @@
...
+import { mainnet } from 'wagmi';
+import { renderWithProviders } from '../../../test';
+import { WalletButton } from './WalletButton';
+
+describe('<WalletButton />', () => {
+  const getWalletButtonLabel = async (connectorId?: string) => {
+    const { findByTestId } = renderWithProviders(
+      <WalletButton wallet={connectorId} />,
+      {
+        chains: [mainnet],
```
KosmosKey, line 12: Unit test just in case 👌

DanielSinclair (approved):

PR #1582, opened 2023-10-25
fix: ethereum.providers support for coinbaseWallet

DanielSinclair:
## Changes
- use standardized `hasInjectedProvider` util to detect Coinbase Wallet more reliably


0xmkh (approved):
Works great 👍

PR #1586, opened 2023-10-29
feat: add tokenaryWallet

grachyov:


DanielSinclair:
@grachyov Seeing CI and deployment errors

packages/rainbowkit/src/wallets/walletConnectors/tokenaryWallet/tokenaryWallet.ts
```diff
@@ -0,0 +1,34 @@
...
+}: TokenaryWalletOptions & InjectedConnectorOptions): Wallet => ({
+  id: 'tokenary',
+  name: 'Tokenary',
+  iconUrl: async () => (await import('./tokenary.svg')).default,
+  iconBackground: '#000000',
+  installed:
+    typeof window !== 'undefined' &&
+    typeof window.ethereum !== 'undefined' &&
+    window.ethereum.isTokenary,
+  hidden: () => isAndroid(),
```
DanielSinclair, line 23: @grachyov Is Tokenary unsupported on Android? Typically we'd still leave the wallet connector visible on the dApp even on unsupported platforms. The exceptions we have thus far are for the `safeWallet` and `injectedWallet` connectors where they are only available in edge cases like the Safe Browser or when an injected wallet connector is not available and the user would need a button to connect their wallet.

grachyov: @DanielSinclair yeah, tokenary is not supported on android. only ios and macos safari.

grachyov: @DanielSinclair what should i do to these deployment errors?

PR #1595, opened 2023-11-01
docs: fix missing en-US localization docs

DanielSinclair:
## Changes
- re-added `localization.mdx` that was inadvertently removed from `en-US` in the last commit
- removed unnecessary `localeDetection` config


0xmkh (approved):
Awesome

PR #1599, opened 2023-11-07
fix: fix popover content is overflow in PropsTable

monodyle:
While researching the documentation of rainbowkit, I identified a small issue related to the UI here.

The problem originates from the data cell `td` element that was a style `white-space: nowrap`. The popover children subsequently inherit this style, causing a UI glitch. So, I have reset the `white-space` within the popover content.

Please find the before and after screenshots below for a better understanding of the changes:

**Before:**
image

**After:**
image


DanielSinclair:
@monodyle Thank you!

DanielSinclair (approved):

PR #1601, opened 2023-11-08
feat: wagmi v2 migration, eip-6963 wallet support

0xmkh:


examples/with-next-app/app/providers.tsx
```diff
@@ -53,28 +50,38 @@ const connectors = connectorsForWallets([
...
-      ledgerWallet({ projectId, chains }),
+      argentWallet({ projectId }),
+      trustWallet({ projectId }),
+      ledgerWallet({ projectId }),
     ],
   },
 ]);
 
 const wagmiConfig = createConfig({
-  autoConnect: true,
```
KosmosKey, line 64: `autoConnect` is deprecated. Users would have to use `useReconnect` hook or specify `reconnectOnMount` prop to `WagmiProvider`. [Link](https://beta.wagmi.sh/react/guides/migrate-from-v1-to-v2#updated-createconfig-parameters)

examples/with-vite/src/main.tsx
```diff
@@ -2,36 +2,51 @@ import './polyfills';
...
 
 const { connectors } = getDefaultWallets({
   appName: 'RainbowKit demo',
   projectId: 'YOUR_PROJECT_ID',
-  chains,
 });
 
 const wagmiConfig = createConfig({
-  autoConnect: true,
+  chains: chains as unknown as readonly [Chain, ...Chain[]],
```
KosmosKey, line 28: Going to fix this but `wagmi` uses tuple types as `readonly [Chain, ...Chain[]]`

packages/example/pages/_app.tsx
```diff
@@ -76,88 +73,89 @@ import {
...
   ...wallets,
   {
     groupName: 'Other',
     wallets: [
-      argentWallet({ chains, projectId }),
-      bifrostWallet({ chains, projectId }),
-      bitgetWallet({ chains, projectId }),
-      bitskiWallet({ chains }),
-      clvWallet({ chains, projectId }),
-      coin98Wallet({ chains, projectId }),
```
KosmosKey, line 124: All the `chains` are now removed. Instead we will give a `CreateConnectorFn` function to wagmi and it will specify all the chains, emitters, events for us.

packages/rainbowkit/src/components/AsyncImage/AsyncImage.tsx
```diff
@@ -62,11 +62,12 @@ export function AsyncImage({
               'src': src,
             }
           : {
-              backgroundSize: 'cover',
+              backgroundSize: 'contain',
```
KosmosKey, line 65: Since `EIP6963` connectors have their own icon it might look a bit over sized if i do `cover`. Just made sure that both rainbowkit connectors and `EIP6963` connectors have background size `contain` so it fills up the space better.

packages/rainbowkit/src/components/ChainModal/Chain.tsx
```diff
@@ -0,0 +1,128 @@
...
+  chainId,
+  currentChainId,
+  switchChain,
+  chainIconSize,
+  isLoading,
+  src,
+  name,
+  iconBackground,
+  idx,
+}: ChainProps) => {
```
KosmosKey, line 34: I created `Chain` component so it will be more readable for us & for other people since `ChainModal` was one of the components that required a good amount of changes. Made sure to seperate components here a bit 👍

packages/rainbowkit/src/components/ChainModal/ChainModal.test.tsx
```diff
@@ -1,21 +1,27 @@
...
-      <ChainModal onClose={() => {}} open />,
+      <ChainModal mockChainId={1} onClose={() => {}} open />,
       {
-        chains: [mainnet], // only supports mainnet
+        chains: [polygon],
         mock: true,
-        mockOptions: { chainId: goerli.id }, // is connected to goerli
       },
     );
+
```
KosmosKey, line 24: wagmi beta doesn't have `chainId` when you do [account mocks](https://beta.wagmi.sh/react/api/connectors/mock#mock). I made sure to pass in a prop called `mockChainId` for now just to mock things and make sure behaviours work as expected.

Probably can rename this `initialChainId`, but seems to work for now.

packages/rainbowkit/src/components/ChainModal/ChainModal.tsx
```diff
@@ -1,33 +1,49 @@
...
+      },
+      onSuccess: () => {
+        if (pendingChainId) setPendingChainId(null);
+      },
+      onError: () => {
+        if (pendingChainId) setPendingChainId(null);
+      },
+      onSettled: () => {
+        onClose();
+      },
```
KosmosKey, line 46: There is no lunger pendingChainId here. Since everything is migrated over to tanstack react query we would have to find out ourselves which chain id is pending. This seems to be working just fine.

packages/rainbowkit/src/components/ConnectButton/formatENS.ts
```diff
@@ -1,4 +1,8 @@
-export function formatENS(name: string): string {
+import { GetEnsNameReturnType } from 'viem';
+
+export function formatENS(name: GetEnsNameReturnType): string {
+  if (!name) return '';
+
```
KosmosKey, line 5: This won't break since a lot of places in code we make sure it's true before formatting it. Just an empty string with early return works fine for now.

image

packages/rainbowkit/src/components/ProfileDetails/ProfileDetails.tsx
```diff
@@ -59,6 +61,7 @@ export function ProfileDetails({
   }
 
   const accountName = ensName ? formatENS(ensName) : formatAddress(address);
+  // @ts-ignore
   const ethBalance = balanceData?.formatted;
```
KosmosKey, line 65: TODO: Wagmi has deprecated the `formatted` value. I'll have to see how we can format balances.

KosmosKey, line 65: TODO: Wagmi has deprecated the formatted value. I'll have to see how we can format balances.

packages/rainbowkit/src/hooks/useMainnetEnsAvatar.ts
```diff
@@ -1,13 +1,11 @@
...
+import { mainnet } from 'wagmi/chains';
 
+export function useMainnetEnsAvatar(name: GetEnsNameReturnType | undefined) {
   const { data: ensAvatar } = useEnsAvatar({
-    chainId,
-    enabled,
-    name,
+    chainId: mainnet.id,
+    name: name ? normalize(name) : undefined,
   });
```
KosmosKey, line 10: If `undefined` the `useEnsAvatar` will be disabled and not make any queries

packages/rainbowkit/src/components/AsyncImage/AsyncImage.tsx
```diff
@@ -62,12 +62,12 @@ export function AsyncImage({
...
             })}
         height="full"
         position="absolute"
         style={{
-          touchCallout: 'none',
+          WebkitTouchCallout: 'none',
```
KosmosKey, line 70: Weird that CI/CD didn't catch this issue, but it somehow catches in this PR. However the `touchCallout` doesn't exist and we have to use `WebkitTouchCallout`.

packages/rainbowkit/src/components/RainbowKitProvider/AuthenticationContext.tsx
```diff
@@ -89,11 +90,11 @@ export function RainbowKitAuthenticationProvider<Message = unknown>({
...
       // Attach the event listener when status is 'authenticated'
-      connector.on('change', handleChangedAccount);
+      connector.emitter.on('change', handleChangedAccount);
 
       // Cleanup function to remove the event listener
       return () => {
-        connector?.off('change', handleChangedAccount);
+        connector.emitter.off('change', handleChangedAccount);
       };
     }
```
KosmosKey, line 99: Took a look at this approach more closely and tried to find some built in functions / hooks for these emitters but seems like it's only built for `onConnect` and `onDisconnect`. For now we can just pull in the `emitter` from the connector.

packages/rainbowkit/src/components/RainbowKitProvider/RainbowKitChainContext.tsx
```diff
@@ -9,10 +8,6 @@ export interface RainbowKitChain {
   iconBackground?: string;
 }
 
-// This type is a combination of wagmi and RainbowKit chain types to make
-// it easier for consumers to define their chain config in a single place.
-export type Chain = WagmiChain & RainbowKitChain;
-
```
KosmosKey, line 15: No longer used. Most of the files which used this type was from [rainbowkit connectors](https://github.com/search?q=repo%3Arainbow-me%2Frainbowkit+%2FRainbowKitChainContext&type=code&p=3)

packages/rainbowkit/src/locales/ru_RU.json
```diff
@@ -86,6 +86,7 @@
     "recommended": "Рекомендуемые",
     "other": "Другие",
     "popular": "Популярные",
+    "installed": "Установлен",
     "more": "Больше",
     "others": "Другие"
   },
```
KosmosKey, line 92: Cool if we could add i18n support for `installed` word for EIP6963

packages/rainbowkit/src/wallets/Wallet.ts
```diff
@@ -7,13 +7,12 @@ export type InstructionStepName =
...
   mobile?: {
-    getUri?: () => Promise<string>;
+    getUri?: (uri: string) => string;
   };
   desktop?: {
-    getUri?: () => Promise<string>;
+    getUri?: (uri: string) => string;
     instructions?: {
       learnMoreUrl: string;
       steps: {
```
KosmosKey, line 18: Reasons for this change & approach we're taking from here:

1. You cannot create the connector first and then prepare to get the URI.
Instead, you would need to pass in a function (`CreateConnectorFn`) to wagmi's `createConfig`.

2. In the React tree, we can ensure that the URI is obtained from the connector. Once we've done that, we can then convert the URI to the format specified in the RainbowKit connector for the `mobile.getUri`, `desktop.getUri` and `qrCode.getUri`.

packages/rainbowkit/src/wallets/Wallet.ts
```diff
@@ -70,23 +70,34 @@ export type Wallet<C extends Connector = Connector> = {
...
-      connector: Connector;
-      installed?: boolean;
-      name: string;
-    }[];
-  }) => boolean;
-  createConnector: () => RainbowKitConnector<C>;
-};
+  hidden?: () => boolean;
+  createConnector: (options: WalletOptionsParams) => CreateConnectorFn;
+} & RainbowKitConnector;
```
KosmosKey, line 75: Wagmi team told me if we want to specify our own download links and custom options we would have to pass it in like this:

```
createConnector((config) => ({
    ...injected(injectedConfig)(config),
    ...walletOptions // Here all our configs (e.g download links, icon url and etc...)
}));
```

packages/rainbowkit/src/wallets/Wallet.ts
```diff
@@ -70,23 +70,34 @@ export type Wallet<C extends Connector = Connector> = {
...
+export type CreateConnector = (walletOptions: {
+  rkDetails: RainbowKitDetails;
+}) => CreateConnectorFn;
+
+// This is the default connector you get at first from wagmi
+// "Connector" + rainbowkit details we inject into the connector
+export type WagmiConnectorInstance = Connector & {
+  rkDetails: RainbowKitDetails;
+};
+
```
KosmosKey, line 100: I recently added `rkDetails` to the connectors. It's not the best, but it was really needed. Skipping `rkDetails` could lead to bugs, like wallet connecting and disconnecting after 5 seconds randomly. This happens because we're overriding the object like this:

```
createConnector((config) => ({
    ...injected(injectedConfig)(config),
    ...walletOptions // If `rkDetails` not passed here it will introduce bugs
}));
```

By implementing `rkDetails`, we avoid these bugs and can put our own icons and download links in the `useWalletConnectors`  hook. This is particularly important since wagmi uses emitters and tanstack. I've also taken a look and without `rkDetails` they will include our values in their functions which is not good.

packages/rainbowkit/src/wallets/connectorsForWallets.test.ts
```diff
@@ -1,236 +1,84 @@
...
 
       expect(connectors.length).toBe(2);
 
-      expect(getWalletInstances(connectors[0]).length).toBe(1);
-      expect(getWalletInstances(connectors[0])[0].index).toBe(0);
-      expect(getWalletInstances(connectors[0])[0].name).toBe(
-        'Test WalletConnect Wallet',
-      );
+      expect(isValidConnector(connectors[0])).toBe(true);
+      expect(isValidConnector(connectors[1])).toBe(true);
```
KosmosKey, line 41: Some of the unit test needed to be refactored due to migration of wagmi v2

packages/rainbowkit/src/wallets/connectorsForWallets.ts
```diff
@@ -1,151 +1,109 @@
...
+      const isWalletConnectConnector = walletMeta.id === 'walletConnect';
+
+      if (isWalletConnectConnector) {
+        connectors.push(
+          createConnector(
+            params({ showQrModal: true, isWalletConnectModalConnector: true }),
+          ),
+        );
+      }
+
```
KosmosKey, line 101: The wallet connect modal also needed to be included. Made a cool trick here to push connector again just with `showQrModal` and then in `useWalletConnectors` hook we can use that connector to open the wallet connect modal.

packages/rainbowkit/src/wallets/getDefaultWallets.ts
```diff
@@ -25,13 +21,12 @@ export const getDefaultWallets = ({
     {
       groupName: 'Popular',
       wallets: [
-        injectedWallet({ chains }),
```
KosmosKey, line 28: I had to get rid of this `injectedWallet` for now. Since we can't see the connector information such as `id`, `name` and etc. It's only visible after `createConfig` from wagmi.

packages/rainbowkit/src/wallets/getInjectedConnector.test.ts
```diff
@@ -1,90 +1,28 @@
...
     const hasRainbow = hasInjectedProvider('isRainbow');
     expect(hasRainbow).toEqual(true);
   });
 
   it('only metamask provider', () => {
-    window.ethereum = { isMetaMask: true } as WindowProvider;
+    window.ethereum = { isMetaMask: true };
     const hasRainbow = hasInjectedProvider('isRainbow');
     expect(hasRainbow).toEqual(false);
   });
```
KosmosKey, line 28: Small cleanups...

packages/rainbowkit/src/wallets/getInjectedConnector.ts
```diff
@@ -41,20 +34,38 @@ function getInjectedProvider(
...
+      : {};
+
+    return createConnector((config) => ({
+      // Spread the injectedConfig object, which may be empty or contain the target function
+      ...injected(injectedConfig)(config),
+      ...walletOptions,
+    }));
+  };
+}
+
```
KosmosKey, line 57: Here is the injected connector which we will be using for all providers 👍

packages/rainbowkit/src/wallets/getWalletConnectConnector.ts
```diff
@@ -1,105 +1,66 @@
...
+): ReturnType<typeof walletConnect> => {
+  const key = `${projectId}_${showQrModal}`;
 
-type WalletConnectConnectorConfig = ConstructorParameters<
-  typeof WalletConnectConnector
->[0];
-export type WalletConnectConnectorOptions =
-  // @ts-ignore
-  WalletConnectConnectorConfig['options'];
+  const sharedWalletConnector = walletConnectInstances.get(key);
```
KosmosKey, line 18: This function will do very similar to what we do [here](https://github.com/rainbow-me/rainbowkit/blob/bd694e8457260cd3dd1c28a2076866d2a61f1d11/packages/rainbowkit/src/wallets/getWalletConnectConnector.ts#L102) just instead we will store it as a function so it can be reusable

packages/rainbowkit/src/wallets/groupedWallets.ts
```diff
@@ -0,0 +1,62 @@
...
+};
+
+export const isEIP6963Connector = (wallet: WalletInstance) => {
+  return !!(
+    !wallet.isRainbowKitConnector &&
+    wallet.icon &&
+    wallet.uid &&
+    wallet.name
+  );
+};
```
KosmosKey, line 34: This file contains all the helper function that we need in order to make rainbowkit connector + eip6963 work fine ✨

packages/rainbowkit/src/wallets/useWalletConnectors.ts
```diff
@@ -1,37 +1,55 @@
...
 }
 
+const eip6963StrPrefix = 'eip6963';
+
 export function useWalletConnectors(): WalletConnector[] {
   const rainbowKitChains = useRainbowKitChains();
   const intialChainId = useInitialChainId();
   const { connectAsync, connectors: defaultConnectors_untyped } = useConnect();
-  const defaultConnectors = defaultConnectors_untyped as Connector[];
+  const defaultCreatedConnectors =
```
KosmosKey, line 44: Some refactoring regarding to [this](https://github.com/rainbow-me/rainbowkit/pull/1601/files#r1394090065)

packages/rainbowkit/src/wallets/useWalletConnectors.ts
```diff
@@ -71,64 +88,118 @@ export function useWalletConnectors(): WalletConnector[] {
...
+  const getWalletConnectUri = async (
+    connector: Connector,
+    uriConverter: (uri: string) => string,
+  ): Promise<string> => {
+    const provider = await connector.getProvider();
+
+    if (connector.id === 'coinbase') {
+      // @ts-expect-error
+      return provider.qrUrl;
+    }
```
KosmosKey, line 100: `provider` is an `unknown` type for now

```
getProvider(parameters?: { chainId?: number | undefined } | undefined): Promise<unknown>
```

packages/rainbowkit/src/wallets/walletConnectors/argentWallet/argentWallet.ts
```diff
@@ -39,51 +18,37 @@ export const argentWallet = ({
...
+        ? uri
+        : `argent://app/wc?uri=${encodeURIComponent(uri)}`;
+    },
+  },
+  qrCode: {
+    getUri: (uri: string) => uri,
+    instructions: {
+      learnMoreUrl: 'https://argent.xyz/learn/what-is-a-crypto-wallet/',
+      steps: [
+        {
```
KosmosKey, line 33: Everything from here and down 👇 will be changes to custom connectors only

examples/with-next-app-i18n/app/[locale]/providers.tsx
```diff
@@ -54,31 +51,48 @@ const connectors = connectorsForWallets([
...
+      argentWallet({ projectId }),
+      trustWallet({ projectId }),
+      ledgerWallet({ projectId }),
     ],
   },
 ]);
 
 const wagmiConfig = createConfig({
-  autoConnect: true,
+  chains: chains as unknown as readonly [Chain, ...Chain[]],
```
Comment (KosmosKey) [inline examples/with-next-app-i18n/app/[locale]/providers.tsx:62]: 👇 [here](https://github.com/rainbow-me/rainbowkit/pull/1601/files#r1393185271)

packages/rainbowkit/src/components/ConnectOptions/MobileOptions.tsx
```diff
@@ -154,7 +143,9 @@ enum MobileWalletStep {
 
 export function MobileOptions({ onClose }: { onClose: () => void }) {
   const titleId = 'rk_connect_title';
-  const wallets = useWalletConnectors();
+  const wallets = useWalletConnectors().filter(
+    (wallet) => wallet.isRainbowKitConnector,
+  );
```
KosmosKey, line 148: We don't want to show EIP6963 connectors here

packages/rainbowkit/src/components/RainbowKitProvider/provideRainbowKitChains.ts
```diff
@@ -144,7 +144,7 @@ export const provideRainbowKitChains = <Chain extends RainbowKitChain>(
     return {
       ...chain,
       name: defaultMetadata.name ?? chain.name, // favor colloquial names
-      iconUrl: chain.iconUrl ?? defaultMetadata.iconUrl,
-      iconBackground: chain.iconBackground ?? defaultMetadata.iconBackground,
+      iconUrl: defaultMetadata.iconUrl,
+      iconBackground: defaultMetadata.iconBackground,
     } as Chain;
```
KosmosKey, line 149: Since we will pass in `chains` from wagmi directly we won't need to have this `iconUrl` and `iconBackground`

examples/with-vite/src/main.tsx
greg-schrammel: how about 

```ts
const chains = [mainnet, polygon, optimism, arbitrum, base, zora] as const;
```
and
```ts
export interface RainbowKitProviderProps {
  chains: ReadonlyArray<RainbowKitChain>;
  ...
```

packages/rainbowkit/src/hooks/useMainnetEnsName.ts
```diff
@@ -1,13 +1,10 @@
...
 
 export function useMainnetEnsName(address: string | undefined) {
-  const { chainId, enabled } = useMainnet();
-
   const { data: ensName } = useEnsName({
-    address,
-    chainId,
-    enabled,
+    chainId: mainnet.id,
+    address: address as `0x`,
```
greg-schrammel, line 7: can't we have the arg type as `Address` instead of casting here?
```diff
export function useMainnetEnsName(
- address: string | undefined
+ address: Address | undefined
) {
```

KosmosKey: Oh yeah we can. I guess we can import it from `viem`, but good catch 🙏 🙌

packages/create-rainbowkit/tsconfig.json
```diff
@@ -1,10 +1,10 @@
 {
   "extends": "../../tsconfig.json",
   "compilerOptions": {
-    "preserveSymlinks": true,
+    "preserveSymlinks": false,
```
KosmosKey, line 4: This is needed for wagmi beta to not get `ts-errors`

packages/rainbow-button/CHANGELOG.md
```diff
@@ -16,17 +16,17 @@
...
     RainbowButtonProvider,
-    RainbowConnector,
+    rainbowConnector,
   } from "@rainbow-me/rainbow-button";
 
   const wagmiClient = createConfig({
-    connectors: [new RainbowConnector({ chains, projectId })],
+    connectors: [rainbowConnector({ projectId })],
     publicClient,
   });
```
KosmosKey, line 32: Wagmi V2 no longer uses classes like `new InjectedConnector()`. Instead they have pure functions like `walletConnect()` or `ledger()`. Want to keep it similar to what they have 🙏

0xartcro:
Since you are doign a big refactor, any chance we could have "switch chain" option available before connecting the wallet?

KosmosKey:
> Since you are doign a big refactor, any chance we could have "switch chain" option available before connecting the wallet?

Hey @0xartcro. Not sure if this is possible since i believe you are required to have your wallet connected first before switching chain. If you are interested in such feature we can do a separate PR for this, but not on this one.

0xartcro:
> > Since you are doign a big refactor, any chance we could have "switch chain" option available before connecting the wallet?
> 
> Hey @0xartcro. Not sure if this is possible since i believe you are required to have your wallet connected first before switching chain. If you are interested in such feature we can do a separate PR for this, but not on this one.

Yeah, that's the problem currently :( I would love to be able to switch chain without connecting the wallet (for read only purpose) so it's easier to implement multi-chain website

KosmosKey:
> > > Since you are doign a big refactor, any chance we could have "switch chain" option available before connecting the wallet?
> > 
> > 
> > Hey @0xartcro. Not sure if this is possible since i believe you are required to have your wallet connected first before switching chain. If you are interested in such feature we can do a separate PR for this, but not on this one.
> 
> Yeah, that's the problem currently :( I would love to be able to switch chain without connecting the wallet (for read only purpose) so it's easier to implement multi-chain website

@0xartcro It seems like it's more of an architectural issue rather than `rainbowkit` issue. You can always managed states within your app and let user select the chain they want to use before opening `rainbowkit` connect modal.  

Then what you can do is use `wagmi` hooks such as `useSwitchNetwork` + `useAccount ` to determine when a user has connected their wallet + switch to the chain that user has selected previously on your app.

.changeset/twelve-tigers-stare.md
```diff
@@ -0,0 +1,274 @@
...
+  },
+});
+```
+
+**EIP6963 Not Supported on Mobile**
+
+Please note that EIP6963 is not supported on mobile devices. Instead, we use the wallets provided to `connectorsForWallets` and/or `getDefaultWallets` and generate a WalletConnect URI to redirect you to the wallet app.
+
+**Rainbow Button Updates**
+
```
DanielSinclair, line 161: We should create a different changeset for that package

.changeset/twelve-tigers-stare.md
```diff
@@ -0,0 +1,274 @@
...
+"with-next-mint-nft": patch
+"@rainbow-me/create-rainbowkit": patch
+"@rainbow-me/rainbow-button": patch
+"with-next-app": patch
+"with-remix": patch
+"@rainbow-me/rainbowkit": patch
+"with-next": patch
+"with-vite": patch
+"example": patch
+"site": patch
```
DanielSinclair, line 20: We should create separate changesets for rainbowkit, rainbow-button, create-rainbowkit, and the examples (that can be all in one)

.changeset/twelve-tigers-stare.md
```diff
@@ -0,0 +1,274 @@
...
+---
+
+**Wagmi V2 Integration and EIP6963 Support**
```
DanielSinclair, line 23: We should separate this into 2 pieces, first focussing on the Migration guidance, then discussing new features.
Check out these: 
- https://github.com/rainbow-me/rainbowkit/releases/tag/%40rainbow-me%2Frainbowkit%401.0.0
- https://github.com/rainbow-me/rainbowkit/releases/tag/%40rainbow-me%2Frainbowkit%400.8.0

.changeset/twelve-tigers-stare.md
```diff
@@ -0,0 +1,274 @@
...
+const App = () => {
+  return (
+    <WagmiConfig config={wagmiConfig}>
+      <RainbowKitProvider chains={chains}>{/* Your App */}</RainbowKitProvider>
+    </WagmiConfig>
+  );
+};
+```
+
+**After:**
```
DanielSinclair, line 65: Checkout this markdown formatting for diffs: https://github.com/wevm/wagmi/releases/tag/wagmi%400.12.0

.changeset/twelve-tigers-stare.md
```diff
@@ -0,0 +1,274 @@
...
+        <RainbowKitProvider chains={chains}>
+          {/* Your App */}
+        </RainbowKitProvider>
+      </QueryClientProvider>
+    </WagmiProvider>
+  );
+};
+```
+
+**Prevent wallet Duplication with EIP6963 in RainbowKit**
```
DanielSinclair, line 104: We could add this section to the Custom Wallet Connectors doc; probably don't need in the migration guide

.changeset/twelve-tigers-stare.md
```diff
@@ -0,0 +1,274 @@
...
+- Frontier Wallet
+- OKX Wallet
+- Phantom
+- Rabby Wallet
+- Talisman
+- TokenPocket
+- XDEFI Wallet
+- Zerion
+
+**Disabling EIP6963 Support in Your dApp**
```
DanielSinclair, line 141: I think we can drop this; we don't want dApps to do this

.changeset/twelve-tigers-stare.md
```diff
@@ -0,0 +1,274 @@
...
+  chains,
+  connectors,
+  multiInjectedProviderDiscovery: false, // <- disable it here
+  transports: {
+    [mainnet.id]: http(),
+  },
+});
+```
+
+**EIP6963 Not Supported on Mobile**
```
DanielSinclair, line 156: We can probably drop this part; should be intuitive

.changeset/twelve-tigers-stare.md
```diff
@@ -0,0 +1,274 @@
...
+    <WagmiProvider config={wagmiConfig}>
+      <QueryClientProvider client={client}>
+        <RainbowButtonProvider>{/* Your App */}</RainbowButtonProvider>
+      </QueryClientProvider>
+    </WagmiProvider>
+  );
+};
+```
+
+**Wallet Button Updates**
```
DanielSinclair, line 234: Are there any breaking changes here? It seems like it might not be necessary to explain

packages/rainbow-button/src/connectors/rainbow.ts
```diff
@@ -1,16 +1,12 @@
...
+  return connector;
 }
 
-function RainbowConnector(options: RainbowConnectorOptions) {
-  const [rainbowConnector] = connectorsForWallets([rainbowWallet(options)]);
-  return rainbowConnector;
-}
-
-export { RainbowConnector };
+export { rainbowConnector };
```
DanielSinclair, line 12: This would be a breaking change so we should add that to the changeset

packages/rainbow-button/src/components/RainbowButton.tsx
```diff
@@ -10,8 +11,8 @@ export function RainbowButtonProvider({
   'chains' | 'avatar' | 'initialChain' | 'modalSize' | 'showRecentTransactions'
 >) {
   return (
-    // chains array is intentionally left empty
-    <RainbowKitProvider chains={[]} {...options}>
+    // Wagmi requires 'chains' as a tuple type with at least one chain
+    <RainbowKitProvider chains={[mainnet]} {...options}>
```
DanielSinclair, line 15: This should only be consumed within RainbowKit, and not passed to Wagmi, right? Do we need to add a chain here?

packages/rainbow-button/package.json
```diff
@@ -39,11 +39,8 @@
   "peerDependencies": {
     "react": ">=17",
     "react-dom": ">=17",
-    "viem": "~0.3.19 || ^1.0.0",
-    "wagmi": "~1.0.1 || ~1.1.0 || ~1.2.0 || ~1.3.0 || ~1.4.0"
-  },
-  "devDependencies": {
-    "@wagmi/core": "^1.4.3"
+    "viem": "2.0.0",
+    "wagmi": "2.0.0-rc.2"
```
DanielSinclair, line 43: This looks problematic; we still need these as peers I believe

packages/rainbow-button/CHANGELOG.md
```diff
@@ -41,17 +41,17 @@
...
   import "@rainbow-me/rainbow-button/styles.css";
   import {
     RainbowButtonProvider,
-    RainbowConnector,
+    rainbowConnector,
   } from "@rainbow-me/rainbow-button";
 
-  const config = createConfig({
-    connectors: [new RainbowConnector({ chains, projectId })],
+  const wagmiClient = createConfig({
```
DanielSinclair, line 53: For styling, should `wagmiClient` here actually be `config`? Referencing this: https://rc.wagmi.sh/react/getting-started#create-config

packages/rainbow-button/CHANGELOG.md
DanielSinclair, line 1: We typically don't want to change the Changelog like this; we'd just want to add to the changeset which then will be amended with the appropriate migration guide

packages/rainbowkit/src/index.ts
DanielSinclair, line 1: I think `getWalletConnectConnector`, `WalletList`, and `Chain` here would be breaking changes for dapps with custom chains or wallets. Was there a goal/reason to remove these?

packages/rainbowkit/package.json
```diff
@@ -44,8 +44,8 @@
   "peerDependencies": {
     "react": ">=17",
     "react-dom": ">=17",
-    "viem": "~0.3.19 || ^1.0.0",
-    "wagmi": "~1.0.1 || ~1.1.0 || ~1.2.0 || ~1.3.0 || ~1.4.0"
+    "viem": "2.0.0",
```
DanielSinclair, line 47: I suppose we should always favor `^2.0.0` for peers and depens

packages/rainbowkit/src/components/ChainModal/Chain.tsx
DanielSinclair: Noice 👌

packages/rainbowkit/src/wallets/getDefaultWallets.ts
```diff
@@ -1,38 +1,33 @@
-import type { Connector } from 'wagmi';
-import type { Chain } from '../components/RainbowKitProvider/RainbowKitChainContext';
-import type { WalletList } from './Wallet';
+import { CreateConnectorFn } from 'wagmi';
```
DanielSinclair, line 1: Is `import type` still viable here? Seeing a few places where we should use that (like the `Chain` type)

packages/example/pages/_app.tsx
```diff
@@ -53,128 +53,111 @@ import {
...
+    [mainnet.id]: http(),
+    [polygon.id]: http(),
+    [optimism.id]: http(),
+    [arbitrum.id]: http(),
+    [base.id]: http(),
+    [zora.id]: http(),
+    [bsc.id]: http(),
+    [zkSync.id]: http(),
+    [goerli.id]: http(),
+    [holesky.id]: http(),
```
DanielSinclair, line 159: We may need to manually add the Alchemy transports here now that `alchemyProvider` is deprecated. (like this: https://rc.wagmi.sh/react/api/transports/fallback#usage) 

At least for our example dapp and docs site. We should remove the mentions of the ALCHEMY API Key env elsewhere

packages/rainbow-button/CHANGELOG.md
KosmosKey: Oops yeah my bad will revert

packages/rainbowkit/src/index.ts
KosmosKey: Reverting it back rn 🙏

I did some changes to the `Chain` type we have in our `RainbowKitChainContext` since wagmi has strict types when it comes to chains and didn't want to mix them together. Going to do this for now :

`export type { RainbowKitChain as Chain } from ...`

packages/rainbow-button/package.json
KosmosKey: Yep that's right. Going to update it on all packages / examples like we have here:
- https://github.com/rainbow-me/rainbowkit/blob/1d602e87d8686dbe1d83e6ba64c38022f99c7681/examples/with-next-app/package.json#L16

- https://github.com/rainbow-me/rainbowkit/blob/1d602e87d8686dbe1d83e6ba64c38022f99c7681/packages/rainbow-button/package.json#L42

KosmosKey: Seems like wagmi requires their `viem` package to be `viem@2.0.0-rc.3`. Here are the [docs](https://rc.wagmi.sh/react/getting-started#manual-installation)

packages/rainbow-button/src/components/RainbowButton.tsx
KosmosKey: Yeah good point. Wagmi v2 uses something called tuple types for their `chains` and i wanted to stick to their types as closely as possible for maintenance purposes. 

They force to use one or more chains: [readonly [Chain, ...Chain[]]](https://rc.wagmi.sh/core/api/createConfig#chains)

I think it's fine for now since this won't impact anything the rainbowkit side. I think having `mainnet` as a default chain is fine for now. Let me know if you have any thoughts 🙏

.changeset/twelve-tigers-stare.md
KosmosKey: Kind of wanted to put it in the `features` section: https://www.rainbowkit.com/docs/introduction#features

If it's not suited there i'll put it on Custom Wallet doc instead 🙏

packages/rainbowkit/src/wallets/walletConnectors/walletConnectWallet/walletConnectWallet.ts
```diff
@@ -1,70 +1,24 @@
...
-  chains: Chain[];
-  version: '1';
-  options?: WalletConnectLegacyConnectorOptions;
-}
 
 export interface WalletConnectWalletOptions {
   projectId: string;
-  chains: Chain[];
-  version?: '2';
-  options?: WalletConnectConnectorOptions;
```
DanielSinclair, line 22: We still need to expose `options` here. I think we should keep `version` too but just mark the `1` as `@deprecated` with a typescript warning, and just remove any version checking/logic

packages/rainbowkit/src/wallets/walletConnectors/walletConnectWallet/walletConnectWallet.test.ts
DanielSinclair, line 1: We should keep this test file; just remove v1 specific tests

packages/rainbowkit/src/wallets/walletConnectors/zerionWallet/zerionWallet.ts
```diff
@@ -1,37 +1,13 @@
...
-  chains: Chain[];
-  walletConnectVersion: '1';
-  walletConnectOptions?: WalletConnectLegacyConnectorOptions;
-}
 
 export interface ZerionWalletOptions {
   projectId: string;
-  chains: Chain[];
-  walletConnectVersion?: '2';
-  walletConnectOptions?: WalletConnectConnectorOptions;
```
DanielSinclair, line 24: We need to keep `walletConnectOptions` in all of these

packages/rainbowkit/src/wallets/walletConnectors/walletConnectWallet/walletConnectWallet.ts
KosmosKey: I think it'll be hard to mark version `1` as deprecated since they no longer expose `WalletConnectLegacyConnectorOptions` type. Will keep the `options` 🙏

packages/rainbowkit/src/wallets/Wallet.ts
```diff
@@ -46,14 +45,16 @@ type RainbowKitConnector<C extends Connector = Connector> = {
...
+export type Wallet = {
   id: string;
   name: string;
+  rdns?: string;
   shortName?: string;
   iconUrl: string | (() => Promise<string>);
   iconAccent?: string;
   iconBackground: string;
   installed?: boolean;
+  ready?: boolean;
```
DanielSinclair, line 57: Do we need this here; shouldn't we be fetching this from the Wagmi connector upstream?

packages/rainbowkit/src/wallets/walletConnectors/walletConnectWallet/walletConnectWallet.ts
DanielSinclair: I suppose we could expose this upstream somewhere alongside a shared `projectId` config, if the WalletConnect Connector is shared. We may want to break that into a follow-up update though.

KosmosKey: Good idea yeah. Let me try that at least.

packages/rainbowkit/src/wallets/Wallet.ts
KosmosKey: Nope not needed. Was trying to add my own `ready` since wagmi no longer supports that. We're using `installed` instead.

packages/rainbowkit/src/components/ChainModal/ChainModal.test.tsx
```diff
@@ -86,65 +71,20 @@ describe('<ChainModal />', () => {
...
-    const modal = renderWithProviders(
-      <ChainModal onClose={() => (onCloseGotCalled = true)} open />,
-      {
-        mock: true,
-        mockOptions: {
-          chainId: mainnet.id,
-          flags: { failSwitchChain: true, isAuthorized: true },
-        },
-      },
-    );
```
KosmosKey, line 100: Had to remove this test since wagmi doesn't provide any failed switch chain flags. Here are more info on their [mock features](https://wagmi.sh/core/api/connectors/mock#features)

packages/rainbowkit/src/components/ConnectOptions/MobileOptions.css.ts
```diff
@@ -24,7 +24,7 @@ export const rotatingBorder = style({
   animation: `${snakeBorder} 1s linear infinite`,
   strokeDasharray: '98 196', // Adjusted based on new perimeter calculation
   fill: 'none',
-  strokeLineCap: 'round',
+  strokeLinecap: 'round',
```
KosmosKey, line 27: Weird didn't do any change here :/

packages/rainbowkit/src/components/ConnectOptions/MobileOptions.tsx
```diff
@@ -67,69 +68,62 @@ export function WalletButton({
...
+    };
+
+    if (id !== wcId) onMobileUri();
+
+    // if the id is "walletConnect" then "showWalletConnectModal" will always be true
+    if (showWalletConnectModal) {
+      showWalletConnectModal();
+      onClose?.();
+      return;
+    }
```
KosmosKey: Good news. Fixed WalletConnect modal for mobile. Currently it's not working on mobile in `main` branch, but made sure to add it here 👌

packages/rainbowkit/src/components/ModalSelection/ModalSelection.tsx
```diff
@@ -88,7 +90,9 @@ export const ModalSelection = ({
           <Box alignItems="center" display="flex" flexDirection="row" gap="12">
             <AsyncImage
               background={iconBackground}
-              {...(isMouseOver ? {} : { borderColor: 'actionButtonBorder' })}
+              {...(!isMouseOver && isRainbowKitConnector
+                ? { borderColor: 'actionButtonBorder' }
+                : {})}
```
KosmosKey, line 95: Needed this since EIP6963 do not want to have a border color. It looks a bit ugly if i do it.

packages/rainbowkit/src/components/RainbowKitProvider/RainbowKitChainContext.tsx
```diff
@@ -1,29 +1,27 @@
...
+export interface RainbowKitChain extends Chain {
   id: number;
-  name?: string;
   iconUrl?: string | (() => Promise<string>) | null;
   iconBackground?: string;
 }
 
-// This type is a combination of wagmi and RainbowKit chain types to make
-// it easier for consumers to define their chain config in a single place.
-export type Chain = WagmiChain & RainbowKitChain;
```
KosmosKey, line 14: Made some cleanups here for better types. Looks better if we're extending types rather than creating another `Chain` type here

packages/rainbowkit/src/wallets/Wallet.ts
```diff
@@ -70,23 +70,38 @@ export type Wallet<C extends Connector = Connector> = {
...
-      id: string;
-      connector: Connector;
-      installed?: boolean;
-      name: string;
-    }[];
-  }) => boolean;
-  createConnector: () => RainbowKitConnector<C>;
-};
+  hidden?: () => boolean;
+  createConnector: (options: WalletDetailsParams) => CreateConnectorFn;
```
KosmosKey: `WalletDetailsParams` are used to inject our `Wallet` details into `wagmi`'s connector. All of our `extensions`, `qrCode` and `mobile` data goes into the `CreateConnectorFn`. 

Later wagmi uses all of the `CreateConnectorFn` we've passed and gives back our data later in the react tree 🙌

packages/rainbowkit/src/wallets/getWalletConnectConnector.ts
```diff
@@ -1,63 +1,105 @@
...
+    ...getOrCreateWalletConnectInstance({
+      projectId,
+      walletConnectParameters,
+      // Used in `connectorsForWallets` to add another
+      // walletConnect wallet into rainbowkit with modal popup option
+      rkDetailsShowQrModal: walletDetails.rkDetails.showQrModal,
+    })(config),
+    ...walletDetails,
+  }));
+}
```
KosmosKey, line 84: 🤩

packages/rainbowkit/src/wallets/walletConnectors/walletConnectWallet/walletConnectWallet.ts
```diff
@@ -1,47 +1,32 @@
...
   projectId: string;
-  chains: Chain[];
-  options?: WalletConnectConnectorOptions;
+  walletConnectParameters?: WalletConnectParameters;
 }
 
+// Used for maintance purposes for `connectorsForWallets` logic
+// wcId = "WalletConnect id"
+export const wcId = 'walletConnect';
+
```
KosmosKey, line 13: Sometimes when we use the same variable for comparing and updating stuff like [here](https://github.com/rainbow-me/rainbowkit/blob/main/packages/rainbowkit/src/components/ConnectOptions/MobileOptions.tsx#L87) and [here](https://github.com/rainbow-me/rainbowkit/blob/main/packages/rainbowkit/src/wallets/connectorsForWallets.ts#L134). Just wanna make sure it's maintainable.

Otherwise if we're going to do more features with `walletConnect` we might introduce some bugs by accident so want to make sure that `id` is being shared across all over the app.

packages/rainbowkit/src/index.ts
```diff
@@ -14,8 +14,13 @@ export {
...
   createAuthenticationAdapter,
 } from './components/RainbowKitProvider/AuthenticationContext';
-export type { Wallet, WalletList } from './wallets/Wallet';
-export type { Chain } from './components/RainbowKitProvider/RainbowKitChainContext';
+export type { RainbowKitChain as Chain } from './components/RainbowKitProvider/RainbowKitChainContext';
+export type {
+  Wallet,
+  WalletList,
+  WalletDetailsParams,
+  RainbowKitWalletConnectParameters,
```
KosmosKey, line 22: Good to expose this into rainbowkit so people can use these types if they want to build their own wallets

packages/rainbowkit/src/wallets/walletConnectors/browserWallet/browserWallet.ts
```diff
@@ -1,38 +1,10 @@
...
   iconUrl: async () => (await import('./browserWallet.svg')).default,
   iconBackground: '#fff',
-  hidden: ({ wallets }) =>
-    wallets.some(
-      (wallet) =>
-        wallet.installed &&
-        wallet.name === wallet.connector.name &&
-        (wallet.connector instanceof InjectedConnector ||
-          wallet.id === 'coinbase'),
-    ),
```
DanielSinclair, line 31: Because `hidden` is removed here, we'll need to do this check in `connectorsForWallets` and append `browserWallet` to the list there (as they were removed from `getDefaultWallets`. We should append `safeWallet` and `braveWallet` in the same way

packages/rainbowkit/src/wallets/walletConnectors/walletConnectWallet/walletConnectWallet.ts
```diff
@@ -1,47 +1,31 @@
...
 export interface WalletConnectWalletOptions {
   projectId: string;
-  chains: Chain[];
-  options?: WalletConnectConnectorOptions;
+  options?: RainbowKitWalletConnectParameters;
 }
 
+// Used for maintance purposes for `connectorsForWallets` logic
+// wcId = "WalletConnect id"
+export const wcId = 'walletConnect';
```
DanielSinclair, line 11: Probably unnecessary; we don't change this. Also creates a bit of a circular dependency

packages/rainbowkit/src/wallets/walletConnectors/walletConnectWallet/walletConnectWallet.ts
```diff
@@ -1,47 +1,31 @@
...
+  options,
+}: WalletConnectWalletOptions): Wallet => {
+  const getUri = (uri: string) => uri;
 
-    const connector = getWalletConnectConnector({
-      chains,
+  return {
+    id: wcId,
+    name: 'WalletConnect',
+    installed: true,
```
DanielSinclair, line 22: `installed` should always be `undefined` for this connector (as it isn't eip-1193), right?

packages/rainbowkit/src/wallets/walletConnectors/browserWallet/browserWallet.ts
KosmosKey: It's actually super difficult to do this check unless we try to use `createConfig` inside `connectorsForWallets`. Maybe that's the the only way, but we can't manually get `wagmi`s connector metadata since it returns `CreatConnectorFn`

KosmosKey: We can probably do this in the react tree to so we can pass in `config.interal._connectors` and then pass into our own function again to see wether we should append `browserWallet`, `safeWallet`, `braveBrowser` and etc...

KosmosKey: Also not sure if we should append it in another category group in connect modal

packages/create-rainbowkit/templates/next-app/pages/_app.tsx
```diff
@@ -12,9 +12,11 @@ import {
   polygon,
   sepolia,
   zora,
+  Chain,
 } from 'wagmi/chains';
+import { getDefaultConfig, RainbowKitProvider } from '@rainbow-me/rainbowkit';
 
-const chains = [
+const chains: readonly [Chain, ...Chain[]] = [
```
DanielSinclair, line 19: Is there a way to surpress these warnings without importing the type? Looks like using Wagmi directly dose:
```
export const config = createConfig({
  chains: [mainnet, sepolia],
  transports: {
    [mainnet.id]: http(),
    [sepolia.id]: http(),
  },
})
```

packages/create-rainbowkit/templates/next-app/pages/_app.tsx
```diff
@@ -24,19 +26,34 @@ const chains = [
   ...(process.env.NEXT_PUBLIC_ENABLE_TESTNETS === 'true' ? [sepolia] : []),
 ];
 
-const wagmiConfig = getDefaultConfig({
+const config = getDefaultConfig({
   appName: 'RainbowKit App',
   projectId: 'YOUR_PROJECT_ID',
+  appDescription: 'RainobwKit Demo App',
+  appUrl: 'https://www.rainbowkit.com',
   chains,
```
DanielSinclair, line 34: Let's make `appDescription` and `appUrl` optional

packages/example/pages/_app.tsx
```diff
@@ -108,55 +115,76 @@ const chains = [
...
-        tahoWallet({ chains }),
-        talismanWallet({ chains }),
-        tokenPocketWallet({ projectId, chains }),
-        tokenaryWallet({ chains }),
-        trustWallet({ projectId, chains }),
-        uniswapWallet({ projectId, chains }),
-        xdefiWallet({ chains }),
-        zealWallet({ chains }),
-        zerionWallet({ projectId, chains }),
+        argentWallet({ projectId }),
```
DanielSinclair, line 154: We'll want this to look like `argentWallet` without passing calling the function or passing the projectId.

DanielSinclair, line 154: We'll want this to look like `argentWallet` without passing calling the function or passing the projectId.

packages/create-rainbowkit/templates/next-app/pages/_app.tsx
```diff
@@ -24,19 +26,32 @@ const chains = [
...
+  transports: {
+    [mainnet.id]: http(),
+    [polygon.id]: http(),
+    [optimism.id]: http(),
+    [arbitrum.id]: http(),
+    [base.id]: http(),
+    [zora.id]: http(),
+    [sepolia.id]: http(),
+  },
```
DanielSinclair, line 41: In this case, the user shouldn't need to pass `transports`. We'd automatically infer and map to `http()` in `getDefaultConfig`

packages/example/pages/_app.tsx
```diff
@@ -85,7 +87,12 @@ const RAINBOW_TERMS = 'https://rainbow.me/terms-of-use';
 const projectId =
   process.env.NEXT_PUBLIC_WALLETCONNECT_PROJECT_ID ?? 'YOUR_PROJECT_ID';
 
-const chains = [
+const { wallets } = getDefaultWallets({
+  appName: 'RainbowKit demo',
```
DanielSinclair, line 91: We drop `appName` param here. User will either call `getDefaultConfig` or `connectorsForWallets` and will need to supply the app metadata there

packages/example/pages/_app.tsx
DanielSinclair: The walletConnectors should just be `Wallet` instead of `() => Wallet types` (and devs could just pass a list that looks like wallets: [argentWallet, rainbowWallet] like chains).

packages/rainbow-button/src/components/RainbowButton.tsx
DanielSinclair, line 15: If `chains` in `RainbowKitProvider` is removed or optional, we can drop this need

site/components/Demo/WalletButton.tsx
```diff
@@ -4,14 +4,15 @@ import {
...
+import { chains } from 'components/Provider/Provider';
 import { useRouter } from 'next/router';
 import React from 'react';
 import { walletButtons } from './WalletButton.css';
 
 export function WalletButtonDemo() {
   const { locale } = useRouter() as { locale: Locale };
   return (
-    <RainbowKitProvider chains={[]} locale={locale}>
+    <RainbowKitProvider chains={chains} locale={locale}>
```
DanielSinclair, line 15: `chains` can be removed like above

packages/rainbowkit/src/config/getDefaultConfig.ts
```diff
@@ -1,15 +1,39 @@
...
   metaMaskWallet,
   rainbowWallet,
   walletConnectWallet,
 } from '../wallets/walletConnectors';
 
+type _chains = readonly [Chain, ...Chain[]];
+
+// Define the '_transports' type as a Record
+// It maps each 'Chain' id to a 'Transport'
+type _transports = Record<_chains[number]['id'], Transport>;
```
DanielSinclair, line 24: Can we infer these types from Wagmi? I imagine they'd break if Wagmi fixes this strong readonly Chains typing

packages/rainbowkit/src/config/getDefaultConfig.ts
```diff
@@ -1,15 +1,39 @@
...
+
+interface GetDefaultConfigParameters {
+  appName: string;
+  appDescription?: string;
+  appUrl?: string;
+  appIcon?: string;
+  chains: readonly [Chain, ...Chain[]];
```
DanielSinclair, line 31: We should use the same inferred type as `_chains` to prevent type drift

packages/rainbowkit/src/config/getDefaultConfig.ts
```diff
@@ -1,15 +1,39 @@
...
+  wallets?: WalletList;
+  projectId: string;
+  transports: _transports;
```
DanielSinclair, line 34: `transports` can be optional, and we just infer an `http()` map with the public provider defaults

packages/rainbowkit/src/wallets/connectorsForWallets.ts
```diff
@@ -1,36 +1,28 @@
...
-export function connectorsForWallets(walletList: WalletList): () => Connector[];
-
 /*
   Assemble a list of low-level connectors for use with the
   `WalletButton` and `RainbowButton` components in custom implementations.
 */
-export function connectorsForWallets(wallets: Wallet[]): Connector[];
+export function connectorsForWallets(
+  wallets: Wallet[] | WalletList,
+): CreateConnectorFn[];
```
DanielSinclair, line 18: I think we should accept `appName`, `appDescription`, and `appUrl` here so that it's passed to the WalletConnect connectors through this flow too

packages/rainbowkit/src/config/getDefaultConfig.ts
KosmosKey: I think they don't expose that type. They do it themselves here: 

image

KosmosKey: Yeah they would break. I think we can just a follow up PR on that if it happens.

packages/create-rainbowkit/templates/next-app/pages/_app.tsx
greg-schrammel: yeah you can do it `as const`
```ts
const chains = [mainnet, sepolia] as const
```

packages/example/pages/_app.tsx
KosmosKey: Tbh i think we can drop this part. We already provide default wallets in `getDefaultConfig`. This would simplify the experience much more.

packages/create-rainbowkit/generated-test-app/pages/_app.tsx
```diff
@@ -0,0 +1,55 @@
...
+  transports: {
+    [mainnet.id]: http(),
+    [polygon.id]: http(),
+    [optimism.id]: http(),
+    [arbitrum.id]: http(),
+    [base.id]: http(),
+    [zora.id]: http(),
+    [sepolia.id]: http(),
+  },
+});
```
DanielSinclair, line 39: Nice! The `transports` piece is the only thing I think we would want to cleanup here; should just assume `http()` internally for the public provider and make `transports` overrides optional

packages/example/pages/_app.tsx
```diff
@@ -85,78 +86,99 @@ const RAINBOW_TERMS = 'https://rainbow.me/terms-of-use';
...
-      ]
-    : []),
-];
-
-const { wallets } = getDefaultWallets({
-  appName: 'RainbowKit Demo',
-  projectId,
-  chains,
-});
+const walletList = getDefaultWalletList();
```
DanielSinclair, line 89: Why'd we create a new API? `getDefaultWallets` would actually return a wallet list

packages/example/pages/_app.tsx
```diff
@@ -85,78 +86,99 @@ const RAINBOW_TERMS = 'https://rainbow.me/terms-of-use';
...
+          polygonMumbai,
+          optimismSepolia,
+          arbitrumSepolia,
+          baseSepolia,
+          zoraSepolia,
+        ]
+      : []),
+  ],
+  transports,
+  walletList: [
```
DanielSinclair, line 143: I would just call this `wallets`, since it mirrors the `connectorsForWallets` API

packages/example/pages/_app.tsx
```diff
@@ -85,78 +86,99 @@ const RAINBOW_TERMS = 'https://rainbow.me/terms-of-use';
...
+        subWallet,
+        tahoWallet,
+        talismanWallet,
+        tokenPocketWallet,
+        tokenaryWallet,
+        trustWallet,
+        uniswapWallet,
+        xdefiWallet,
+        zealWallet,
+        zerionWallet,
```
DanielSinclair, line 181: Nice 👌

packages/example/pages/_app.tsx
DanielSinclair: The `WalletList` type is generally a bit confusing; I think we'll want to refactor that in the future. `wallets` seems a bit more intuitive as the external prop, even though it is `WalletList` type under the hood

packages/rainbowkit/src/wallets/getWalletConnectConnector.ts
```diff
@@ -1,67 +1,106 @@
...
-  // We warn developers against using this projectId in their dApp in production.
-  const exampleProjectId = '21fef48091f12692cad574a6f7753643';
-  if (!projectId || projectId === '')
+  walletConnectParameters,
+}: GetWalletConnectConnectorParams): CreateConnector {
+  // Check if the projectId is provided; if not, throw an error.
+  if (!projectId) {
     throw new Error(
       'No projectId found. Every dApp must now provide a WalletConnect Cloud projectId to enable WalletConnect v2 https://www.rainbowkit.com/docs/installation#configure',
     );
```
DanielSinclair, line 96: This is causing issues with the examples; we need to include these fallback projectIds, as documented.

packages/rainbowkit/src/config/getDefaultConfig.ts
```diff
@@ -1,68 +1,93 @@
...
+  appName,
+  appDescription,
+  appUrl,
+  appIcon,
+  wallets,
+  projectId,
+  chains,
+  multiInjectedProviderDiscovery = true,
+  transports,
+}: GetDefaultConfigParameters): WagmiProviderProps['config'] => {
```
DanielSinclair, line 53: Is `WagmiProviderProps['config']` the right return type here?

KosmosKey: Yeah. You can see it also in `<WagmiProvider />`
image

PR #1603, opened 2023-11-10
feat: add decent mint items

cdsiren:
This PR updates the `with-next-mint-nft` examples repository to include Decent's hooks for cross-chain minting.

Updates include:
- `DecentMint`: a new mint button component that prepares calldata for a crosschain transaction based on the `actionArgs` included in the `index.tsx` file.  This component sends the transaction & then tracks its progress, surfacing blockscanner links for users.
- `TokenSelectorUsage`: (optional) this is the Token Selector component from Decent's UI components library.  It lets users select which token they would like to pay with on the chain to which they are connected.  This can be removed if we just want to demonstrate cross-chain tx's using native gas tokens only.  TODO: there seems to be a potential class conflict that is causing some formatting issues in the modal -- functionally works though.
- two new packages: `@decent.xyz/box-hooks` & `@decent.xyz/box-ui`.

Testnet liquidity for cross-chain tx's is extremely poor, so we created a new NFT to test with - [block explorer link](https://optimistic.etherscan.io/address/0x3146975BFCCAE722F802BC0Cd540dB1e6c178D1F#readContract).


DanielSinclair:
@cdsiren Bumped this example to separate directory `with-next-decent-box` and used static versioning so that if there are breaking changes it won't fail the CI. Do you have a rate limited example API key we can use for the example? The sign-up process right now is a bit frictioned.

Issue #1557, opened 2023-10-11
[bug] Horizontal scrollbar on Switch Network.


0xCatalyst:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### RainbowKit Version

1.1.1

### wagmi Version

1.4.2

### Current Behavior

When you try to switch networks, a transition effect is triggered by clicking on another network. This effect on the latest versions is causing a horizontal zscrollbar to briefly appear at the bottom of the wrapper for half a second before disappearing.

![brave_CT6Y5aEDui](https://github.com/rainbow-me/rainbowkit/assets/27161969/b738a99e-bd78-4ecd-9172-60fd17ad3dfc)

### Expected Behavior

A horizontal scrollbar should not appear.

### Steps To Reproduce

1. Visit [RainbowKit](https://www.rainbowkit.com/)
2. Connect your wallet.
3. Click on the network to open the "Switch Network" menu.
4. Select any network to switch to and observe the bottom of the "Switch Network" menu.

### Link to Minimal Reproducible Example (CodeSandbox, StackBlitz, etc.)

https://www.rainbowkit.com/

### Anything else?

This behavior wasn't present in **^1.0.11**. Today, I updated to **^1.1.1** and noticed the bug.

I've noticed styling changes in this menu, which might be leading to this bug.

### Version ^1.0.11
![brave_Ike8iD85JR](https://github.com/rainbow-me/rainbowkit/assets/27161969/e88f2e04-30cb-45b7-b8b0-3dc6d69ff9a0)
![brave_bcVHfoJpOZ](https://github.com/rainbow-me/rainbowkit/assets/27161969/bbbc2289-955d-429a-97ad-1368fbf150d1)

---

### Version ^1.1.1
![brave_brWhtQ9WhZ](https://github.com/rainbow-me/rainbowkit/assets/27161969/84791be0-f539-4008-b6f7-2c1e88ab146b)
![brave_BbgTP2guF3](https://github.com/rainbow-me/rainbowkit/assets/27161969/9c02ec7a-6590-45a2-830b-4656d72fc4bc)

KosmosKey:
@0xCatalyst I can't see any overflows happening in half a second when switching the chain. Can you update to the latest version `1.2.0` ? It might be related to this issue, but that one is vertical one instead of horizontal ([issue](https://github.com/rainbow-me/rainbowkit/pull/998))

[...]

PR #1607, opened 2023-11-13
fix: chain modal overflows when button is clicked

0xmkh:
Releated to this issue: https://github.com/rainbow-me/rainbowkit/issues/1557


DanielSinclair (approved):

DanielSinclair:
Added one more tweak to prevent this list overflow when 10 chains are added; we had a slight padding and list height miscalculation
Screenshot 2023-11-16 at 12 57 54 PM

Issue #1488, opened 2023-08-29
`eth_getBalance` is requested when showBalance is disabled


DanielSinclair:
### Discussed in https://github.com/rainbow-me/rainbowkit/discussions/1479

<div type='discussions-op-text'>

<sup>Originally posted by **einaralex** August 23, 2023</sup>
My dapp does not need the balance of the native currency and I've made sure that `showBalance` on the ConnectButton is off, but RainbowKitProvider still calls `eth_getBalance` every time it's rendered, is there a way to stop this?</div>

KosmosKey:
We will fix this in this PR: https://github.com/rainbow-me/rainbowkit/pull/1608

I'll mark it as solved for now and should be ready in the next release

DanielSinclair:
This was fixed in #1681 with RainbowKit [1.3.3](https://github.com/rainbow-me/rainbowkit/releases/tag/%40rainbow-me%2Frainbowkit%401.3.3)

PR #1608, opened 2023-11-13
fix: unexpected eth_getBalance calls

0xmkh:
releated to this: https://github.com/rainbow-me/rainbowkit/issues/1488

### DIFF
--- a/packages/rainbowkit/src/components/AccountModal/AccountModal.tsx
+++ b/packages/rainbowkit/src/components/AccountModal/AccountModal.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import { useAccount, useBalance, useDisconnect } from 'wagmi';
 import { useMainnetEnsAvatar } from '../../hooks/useMainnetEnsAvatar';
 import { useMainnetEnsName } from '../../hooks/useMainnetEnsName';
@@ -13,10 +13,11 @@ export interface AccountModalProps {
 
 export function AccountModal({ onClose, open }: AccountModalProps) {
   const { address } = useAccount();
-  const { data: balanceData } = useBalance({ address });
   const ensName = useMainnetEnsName(address);
   const ensAvatar = useMainnetEnsAvatar(ensName);
   const { disconnect } = useDisconnect();
+  const [balanceData, setBalanceData] =
+    useState<ReturnType<typeof useBalance>['data']>();
 
   if (!address) {
     return null;
@@ -36,6 +37,7 @@ export function AccountModal({ onClose, open }: AccountModalProps) {
               ensName={ensName}
               onClose={onClose}
               onDisconnect={disconnect}
+              setBalanceData={setBalanceData}
             />
           </DialogContent>
         </Dialog>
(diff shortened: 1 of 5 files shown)


packages/rainbowkit/src/components/ConnectButton/ConnectButton.tsx
```diff
@@ -43,8 +43,13 @@ export function ConnectButton({
...
   const i18n = useContext(I18nContext);
 
+  const shouldShowBalance =
+    normalizeResponsiveValue(showBalance)[
+      isMobile() ? 'smallScreen' : 'largeScreen'
+    ];
+
   return (
-    <ConnectButtonRenderer>
+    <ConnectButtonRenderer showBalance={shouldShowBalance}>
```
KosmosKey, line 52: I added an extra prop to the `ConnectButtonRendered` component called `showBalance`. We can pass either `true` or `false` and it will make sure to fetch the balance only if it's `true`.

packages/rainbowkit/src/components/AccountModal/AccountModal.tsx
```diff
@@ -13,10 +13,11 @@ export interface AccountModalProps {
 
 export function AccountModal({ onClose, open }: AccountModalProps) {
   const { address } = useAccount();
-  const { data: balanceData } = useBalance({ address });
   const ensName = useMainnetEnsName(address);
   const ensAvatar = useMainnetEnsAvatar(ensName);
   const { disconnect } = useDisconnect();
+  const [balanceData, setBalanceData] =
+    useState<ReturnType<typeof useBalance>['data']>();
```
KosmosKey, line 20: Not sure if it makes sense to do another PR for this, but the chain modal would also call `"eth_getBalance"` randomly. I made sure to only call `"eth_getBalance"` if the chain modal is open.

packages/rainbowkit/src/components/ProfileDetails/ProfileDetails.tsx
```diff
@@ -32,10 +33,21 @@ export function ProfileDetails({
...
+  const { data: wagmiBalanceData } = useBalance({
+    address,
+    enabled: !balanceData,
+  });
+
+  // biome-ignore lint/nursery/useExhaustiveDependencies: <explanation>
+  useEffect(() => {
+    if (wagmiBalanceData) setBalanceData(wagmiBalanceData);
+  }, [wagmiBalanceData?.value]);
+
```
DanielSinclair, line 50: Does the `balanceData` prop that's passed down here trigger the same behavior upon re-render?

packages/rainbowkit/src/components/ConnectButton/ConnectButton.tsx
DanielSinclair: I think we'll want to avoid this addition since it exposes a new prop to the `ConnectButton.Custom` API

DanielSinclair: Not sure the best way to handle this. I wonder if we should create something similar to `ShowRecentTransactionsContext` that's exposed in `RainbowKitProvider` to requested limited RPC calls. That seems like the root cause of the issue; some dApps don't want to pay for excessive RPC usage.

jxom:
Can you not [set `staleTime`](https://wagmi.sh/react/hooks/useBalance#staletime-optional) instead of doing these workarounds?

Issue #934, opened 2022-12-29
[bug] Excessive CPU usage from invisible spinning svg


0xTARC:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### RainbowKit Version

0.8.1

### wagmi Version

0.9.0

### Current Behavior

Invisible spinning SVG causing ridiculous amount of CPU usage

I think caused by this PR: https://github.com/rainbow-me/rainbowkit/commit/d2a2a6aead377fcab420903377e65eb9f1e00e9f

![Recording 2022-12-28 at 20 54 18](https://user-images.githubusercontent.com/97852365/209901315-cd95c474-06e7-4abd-ad68-ffd89678cfdb.gif)

I would have fixed it myself but couldn't get the repo to build 

see it live here: https://jpoold-ui.vercel.app/
code: https://github.com/0xTARC/jpoold-ui

### Expected Behavior

That spinning SVG shouldn't be there

### Steps To Reproduce

Install rainbow kit

Connect wallet

Observe spinning svg or super high CPU usage in chrome task manager

0xTARC:
we faced this issue in fiat dao frontend as well and fixed it with a simple css override
https://github.com/fiatdao/fiat-ui-i/commit/ab965a637211a8b6ca6b46f789d0020ec4603799

DanielSinclair:
A weird one! Don't quite understand why the `SpinnerIcon` styling would be leaking in this way 🤔

KosmosKey:
Hey @0xTARC. I have created a pull request for this bug: https://github.com/rainbow-me/rainbowkit/pull/1609. It will be merged on one of the next releases.

PR #1609, opened 2023-11-13
fix: excessive cpu usage from avatar component

0xmkh:
Related to this issue: https://github.com/rainbow-me/rainbowkit/issues/934

### DIFF
--- a/packages/rainbowkit/src/components/Avatar/Avatar.tsx
+++ b/packages/rainbowkit/src/components/Avatar/Avatar.tsx
@@ -45,7 +45,7 @@ export function Avatar({ address, imageUrl, loading, size }: AvatarProps) {
       >
         <AvatarComponent address={address} ensImage={imageUrl} size={size} />
       </Box>
-      {typeof loading === 'boolean' && (
+      {loading && (
         <Box
           color="accentColor"
           display="flex"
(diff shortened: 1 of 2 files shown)


packages/rainbowkit/src/components/Avatar/Avatar.tsx
```diff
@@ -45,7 +45,7 @@ export function Avatar({ address, imageUrl, loading, size }: AvatarProps) {
       >
         <AvatarComponent address={address} ensImage={imageUrl} size={size} />
       </Box>
-      {typeof loading === 'boolean' && (
+      {loading && (
```
KosmosKey, line 48: Make sure to only show the spinner if it's `true`. Since `typeof loading === 'boolean'` will always be true because it was always a boolean. It had `false` always.

nickbytes (approved):

packages/rainbowkit/src/components/Avatar/Avatar.tsx
```diff
@@ -45,17 +45,12 @@ export function Avatar({ address, imageUrl, loading, size }: AvatarProps) {
...
         <Box
           color="accentColor"
           display="flex"
           height="full"
           position="absolute"
-          style={{
```
nickbytes, line 54: You need to add these styles back, or no?

KosmosKey: Ah yeah. They are not needed anymore since they are only used for transition effects

DanielSinclair (approved):
Nice 👌

PR #1618, opened 2023-11-20
chore: next-auth upgrade, peer version compatibility for v5

DanielSinclair:
## Changes
- upgraded `next-auth` in examples to `4.24.5`
- added strict peer incompatibility with v5


packages/example/pages/api/auth/[...nextauth].ts
```diff
@@ -44,6 +44,7 @@ export function getAuthOptions(req: IncomingMessage): NextAuthOptions {
             id: siwe.address,
           };
         } catch (e) {
+          console.error('siwe authorization failed', e);
```
Comment (KosmosKey) [inline packages/example/pages/api/auth/[...nextauth].ts:47]: Is this needed ?

0xmkh (approved):
Looks good 👍

packages/example/pages/api/auth/[...nextauth].ts
Comment (DanielSinclair) [reply packages/example/pages/api/auth/[...nextauth].ts:47]: Noticed we weren't catching any logs in case the server-side verification failed, so I think this will be useful in the future; only in the example dapp.

PR #1620, opened 2023-11-20
chore: upgrade biome linter

DanielSinclair:
# Changes
- upgrade `@biomejs/biome` to `1.3.3`
- replace `indentSize` with `indentWidth`
- ignore `lint/complexity/noForEach`
- ignore `lint/performance/noAccumulatingSpread`
- remove deprecated `lint/nursery/useExhaustiveDependencies` ignores
- refactored `useIsMounted` and aligned between site/rainbowkit utils
- removed `mounted` check for `Connect Wallet` button string replacement (caused i18n connectbutton tests to fail after refactor)


KosmosKey:
Are there anyways to not include whole functions and objects like `setIsOpen` or `router` in the useEffect dependency array ? Only use what's needed ?

DanielSinclair:
@KosmosKey What do you mean?

packages/rainbowkit/src/components/ConnectButton/ConnectButton.tsx
```diff
@@ -269,7 +269,7 @@ export function ConnectButton({
                 transition="default"
                 type="button"
               >
-                {mounted && label === 'Connect Wallet'
```
DanielSinclair, line 272: Was this necessary? Tests started failing with the refactored `useIsMounted` hook

site/lib/useIsMounted.ts
DanielSinclair, line 1: Borrowed this from `packages/rainbowkit` for consistency

packages/rainbowkit/src/components/ConnectButton/ConnectButton.tsx
KosmosKey: Aha yes i remember why this change was made. If we remove this then there will be a NextJS hydration error since on server side HTML with i18n is different than what's in the client side. 

On SSR it might be "Connect Wallet" and then if user uses spanish it will be "Conectar la billetera" which will throw hydration error.

KosmosKey: @DanielSinclair It's fine for now since i don't see any hydration error on my side at least. Great job on the `useIsMounted` refactoring 🙏

0xmkh (approved):

0xmkh (approved):

0xmkh (approved):

PR #1638, opened 2023-11-29
wip: bundle only used locales

greg-schrammel:
current bundle size (always) / optimal bundle with single lang (like a ssr nextjs page)
Screenshot 2023-11-29 at 12 36 52
Screenshot 2023-11-29 at 12 36 18

the way I could make it work as expected is using a new entry point for each lang
some info on why barrel files are a pain here https://vercel.com/blog/how-we-optimized-package-imports-in-next-js
```ts
// imports just pt_BR messages, (besides en_US that's the default and used as fallback) 
import pt_BR from '@rainbow-me/rainbowkit/locales/pt_BR'; // (~12kb gzipped)
// imports all translations 
import all from '@rainbow-me/rainbowkit/locales'; // (~200kb gzipped)
```

I think this `RainbowKitProvider.i18n.store` api deserves some love I did it this way test quickly
another option I can think is a `messages` prop in `RainbowKitProvider` that when undefined lazy loads all locales
```ts
// after importing the locales you wish to use, store then in the provider i18n obj
RainbowKitProvider.i18n.store(all); 
// or
RainbowKitProvider.i18n.store({ 'pt-BR': pt_BR });
```

### DIFF
--- a/packages/example/pages/_app.tsx
+++ b/packages/example/pages/_app.tsx
@@ -12,6 +12,14 @@ import {
   lightTheme,
   midnightTheme,
 } from '@rainbow-me/rainbowkit';
+
+// import pt_BR from '@rainbow-me/rainbowkit/locales/pt_BR';
+import all from '@rainbow-me/rainbowkit/locales';
+
+// pt_BR.connect.label = 'Sign in';
+
+RainbowKitProvider.i18n.store(all);
+
 import {
   GetSiweMessageOptions,
   RainbowKitSiweNextAuthProvider,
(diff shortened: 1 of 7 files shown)


packages/example/pages/_app.tsx
```diff
@@ -12,6 +12,14 @@ import {
   lightTheme,
   midnightTheme,
 } from '@rainbow-me/rainbowkit';
+
+// import pt_BR from '@rainbow-me/rainbowkit/locales/pt_BR';
+import all from '@rainbow-me/rainbowkit/locales';
+
+// pt_BR.connect.label = 'Sign in';
```
greg-schrammel, line 19: I like this idea of users been able to update the messages however they like

packages/rainbowkit/locales/package.json
```diff
@@ -0,0 +1,4 @@
+{
+  "main": "../dist/locales/index.js",
+  "sideEffects": false
```
greg-schrammel, line 3: seems like `sideEffects` false does nothing on external packages, so can probably be removed

DanielSinclair:
Closing in favor of #1646, but will try to include some of these changes in the near future

PR #1666, opened 2023-12-19
initial lukso wallet

Heisenburgirs:
This is initial commit for adding Lukso's injected wallet "Universal Profile".

I'm not sure if I've everything correct, let me know which existing wallet fits the category so I can follow it accordingly.

It is an injected wallet that lives in window.lukso


DanielSinclair:
@Heisenburgirs Am seeing CI issues. Please also replicate connectors exactly, including adding strings from our en_US locale.

Heisenburgirs:
I think that should do it.

Heisenburgirs:
Oeps, my bad.

DanielSinclair:
@Heisenburgirs You should aim to get the Lukso chain added to the `viem/chains` package (proxied by `wagmi/chains`) here: https://github.com/wevm/viem/tree/main/src/chains

DanielSinclair:
Will close for now because the CI tests are still failing and our connector format has changed with RainbowKit v2. If the wallet supports EIP-6963, RainbowKit v2 will detect and support the wallet automatically. Reference the test dApp and docs here: https://eip6963.org/

PR #1697, opened 2024-01-11
docs: v2 migration guides, docs, changesets

DanielSinclair:


site/data/en-US/docs/migration-guide.mdx
```diff
@@ -6,6 +6,316 @@ title: Migration Guide
...
++      })(config),
++      ...walletDetails,
++    }));
++  },
+});
+```
+
+**5. Custom chains**
+
+We no longer support custom chains. You no longer can pass `chains` into `<RainbowKitProvider>`
```
Songkeys, line 316: I think this should be stated more clearly - we can still provide custom chains via wagmi config while there is no need to pass into here.

Also I found that custom`iconUrl` is no longer supported?

oxSaturn: Yeah, just curious, without `iconUrl`, how will we include icons for chains that's are not built-in in rainbowkit?

CleanShot-AFdaZkqR@2x

ITZSHOAIB (approved):

site/data/en-US/docs/migration-guide.mdx
KosmosKey: Good catch! We'll introduce a custom api `getChains` to provide custom background color and icon url

site/data/en-US/docs/migration-guide.mdx
```diff
@@ -6,6 +6,373 @@ title: Migration Guide
...
+})
+
++ const queryClient = new QueryClient()
+
+const App = () => {
+  return (
+-    <WagmiConfig config={config}>
++    <WagmiProvider config={config}>
++      <QueryClientProvider client={queryClient}>
+        <RainbowKitProvider chains={chains}>
```
Songkeys, line 68: There is no `chains` arg now, isn't it?

songkeys (approved):

PR #1704, opened 2024-01-13
refactor: `useWindowSize` hook to re-render only after resize completion

### LINKED ISSUE #1440: [bug] RainbowKit Provider re-renders too many times when window is being resized

### RainbowKit Version

1.0.7

### wagmi Version

1.3.9

### Current Behavior

If you wrap your entire App in a RainbowKit Provider it seems that the provider uses a custom `useWindowSize` hook. [Hook Code](https://github.com/rainbow-me/rainbowkit/blob/main/packages/rainbowkit/src/hooks/useWindowSize.ts#L13) 

The hook attaches a `onResize` listener on the window object - I'm assuming in order to render the correct modals for phones/tablets. 
While this behaviour is expected - the listener is not throttled and will re-render multiple times when resizing the browser window on a desktop. 
If the RainbowKit Provider is at the top level of the app - this means that the entire app will re-render multiple times when resizing. 

### Expected Behavior

As a user of RainbowKit I'd expect one of the following:

1. The onResize event handler is throttled (250-300ms) and by trailing event - so that it only re-renders the child elements of the provider when the user has finished resizing the browser. 
2. Less intuitive - but the provider might want to expose a prop/config option that allows dev to directly pass in the current windows size and update it as they fit. 

### Steps To Reproduce

1. Create a new Rainbowkit App - or better yet clone the [CRA example](https://github.com/rainbow-me/rainbowkit/tree/main/examples/with-create-react-app)
2. In the main child of the RainbowKit provider - add a console.log to monitor re-renders. 
3. Start resizing the browser window and notice how many times the app re-renders. 

_No response_

### Anything else?

_No response_

0xmkh:
## Changes
- Updated `useWindowSize` hook to prevent component re-renders on every resize event.
- Implemented a `debounce` function for managing the debounce effect.
- Configured the `debounce` duration to 500 milliseconds for `useWindowSize` hook.
- Forced `min-width` to be `720px` for wide connect modal to avoid ui resizing issues when doing window resize event.

### DIFF
--- a/packages/rainbowkit/src/components/ConnectOptions/DesktopOptions.css.ts
+++ b/packages/rainbowkit/src/components/ConnectOptions/DesktopOptions.css.ts
@@ -1,5 +1,5 @@
 import { style } from '@vanilla-extract/css';
-import { largeScreenMinWidth, sprinkles } from '../../css/sprinkles.css';
+import { sprinkles } from '../../css/sprinkles.css';
 export const QRCodeBackgroundClassName = style([
   {
     background: 'white',
@@ -18,12 +18,7 @@ export const ScrollClassName = style([
 
 // biome-ignore format: design system keys
 export const sidebar = style({
-  '@media': {
-    [`screen and (min-width: ${largeScreenMinWidth}px)`]: {
-      minWidth: '287px',
-    },
-  },
-  'minWidth': '246px',
+  minWidth: '287px',
 });
 
 export const sidebarCompactMode = style({
--- a/packages/rainbowkit/src/components/Dialog/DialogContent.css.ts
+++ b/packages/rainbowkit/src/components/Dialog/DialogContent.css.ts
@@ -42,14 +42,8 @@ export const dialogContentWideMobile = style([
 export const dialogContentWideDesktop = style([
   dialogContent,
   {
-    width: largeScreenMinWidth,
-  },
-  {
-    '@media': {
-      [`screen and (min-width: ${largeScreenMinWidth}px)`]: {
-        width: '720px',
-      },
-    },
+    minWidth: '720px',
+    width: '720px',
   },
 ]);
 
--- a/.changeset/curly-spoons-look.md
+++ b/.changeset/curly-spoons-look.md
@@ -3,3 +3,5 @@
 ---
 
 Refactored `useWindowSize` hook, which previously caused components to re-render on every resize event, to now make components re-render only once the user has finished resizing the window.
+
+Forced `min-width` for wide connect modal to avoid weird ui resizing.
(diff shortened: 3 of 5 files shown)


DanielSinclair:
I wonder if we should memoize `props.children` in `RainbowKitProvider` (or similar). Any time we re-size the window, switch locale, or throw a modal, I think we might re-render the pass-through `children` which would an entire DOM tree.

The debounce is okay to optimize the RainbowKit modal re-renders. We do see this ugly flash resizing below, so I think we need to force a `min-width` to prevent the text resizing. Users traditionally wouldn't see this because the window re-size caught them before the CSS resize
Screenshot 2024-01-12 at 11 06 15 PM

packages/rainbowkit/src/components/ConnectOptions/DesktopOptions.css.ts
```diff
@@ -18,12 +18,7 @@ export const ScrollClassName = style([
 
 // biome-ignore format: design system keys
 export const sidebar = style({
-  '@media': {
-    [`screen and (min-width: ${largeScreenMinWidth}px)`]: {
-      minWidth: '287px',
-    },
-  },
-  'minWidth': '246px',
+  minWidth: '287px',
```
KosmosKey, line 21: Sidebar has weird `minWidth` change from `246px` to `287px` so made sure it's `287px` only 👍

packages/rainbowkit/src/components/Dialog/DialogContent.css.ts
```diff
@@ -42,14 +42,8 @@ export const dialogContentWideMobile = style([
...
-    width: largeScreenMinWidth,
-  },
-  {
-    '@media': {
-      [`screen and (min-width: ${largeScreenMinWidth}px)`]: {
-        width: '720px',
-      },
-    },
+    minWidth: '720px',
+    width: '720px',
```
KosmosKey, line 46: Should be very similar to how we do with [compact modal style](https://github.com/rainbow-me/rainbowkit/blob/main/packages/rainbowkit/src/components/Dialog/DialogContent.css.ts#L58-L61)

KosmosKey:
@DanielSinclair I like this idea. Do you think it'll be worth maybe memoizing our own components like `<ModalProvider />`, `<WalletButtonProvider />` and etc instead of `props.children` ?

I think if we do it with `props.children` it might include a lot of memory usage since there might be huge apps built on top of us which can include heavy logic.

I'll take a look and plan out more how to improve the performance with memoizing, but if we can get at least the window resizing to not re-render many times at once then it'll be a small performance boost + we can test and plan the whole performance boost in another PR 🙌

DanielSinclair (approved):

PR #1706, opened 2024-01-14
chore: format

KosmosKey:
## Changes
- Format files


DanielSinclair:
@KosmosKey I think you might have an issue with your Biome config; these are the recent Biome changes. Ensure you are running the latest Biome extension

PR #1712, opened 2024-01-17
fix: `connector.emitter.on` is not a function

0xmkh:
## Changes
- `connector.emitter` renders partially during page load. Made sure to check if the `emitter` has an `on` event before proceeding.

Example from wagmi's boilerplate app.

https://github.com/rainbow-me/rainbowkit/assets/53529533/12577182-b8dd-4a81-bab2-ee61c01bd0ae

## What to test
- Open your brave browser
- Lock your "brave wallet" if you have one
- Have metamask installed and rabby wallet installed, but make sure rabby wallet is disabled in the "extensions"
- Connect your wallet with the "injected" connector instead of eip6963 one and refresh the page
- If it doesn't crash you're good 👍


DanielSinclair (approved):
Looks good to me!

PR #1717, opened 2024-01-19
feat: realtime balance fetcher for recent transactions

### LINKED ISSUE #1290: [bug] need to handle refresh the balance when any transaction is completed

- [x] I have searched the existing issues

### RainbowKit Version

0.8.0

### wagmi Version

0.11.4

### Current Behavior

i have added ConnectButton in Navbar using react app, navbar render only once when app is started,
what happen whenever  i'm doing any transaction using MetaMask balance is not updating inside ConnectButton component  

### Expected Behavior

you have to re-render that component whenever any changes happen in MetaMask wallet

### Steps To Reproduce

_No response_

_No response_

### Anything else?

_No response_

0xmkh:
## Changes
- Added `useRefetchBalance` hook to `TransactionStoreContext` to support real-time balance support
 
https://github.com/rainbow-me/rainbowkit/assets/53529533/f83a2463-b16d-4674-b7c1-d34ed782ce03

## What to test
1. Replace `address` and `value` [here](https://github.com/rainbow-me/rainbowkit/blob/main/packages/example/pages/index.tsx#L291-L293) and try sending the transaction to yourself
2. After that's done make sure you go to our example dApp [here](https://rainbowkit-example.vercel.app/) and under "Add recent transactions" make sure you're adding the transaction hash and description
3. Wait a few seconds and see your balance being refreshed for connect button and account modal


packages/rainbowkit/src/index.ts
```diff
@@ -29,4 +29,5 @@ export { darkTheme } from './themes/darkTheme';
 export { midnightTheme } from './themes/midnightTheme';
 export { cssStringFromTheme } from './css/cssStringFromTheme';
 export { cssObjectFromTheme } from './css/cssObjectFromTheme';
+export { useRefetchBalance } from './hooks/useRefetchBalance';
```
DanielSinclair, line 32: For these optional hooks, I think we might want to add to create a sub-package like `rainbowkit/hooks` akin to the `/wallets` package.json that we have. That way they get built separate and there is code-splitting in the bundle. I think that will be a helpful paradigm going forward as we enhance other features too

DanielSinclair (changes requested):
Looks like we also need to add docs. Want to create a section below "Advanced" called "Hooks", bump the Modal Hooks doc there as "Modals", and create a new doc for "Balances"?

DanielSinclair:
Noticed some merge conflicts too with v2

DanielSinclair:
I think we could simplify this:
- If the dev uses `useBalance` within their dApp (or calls the `refetch` fn from the return type), then it should also re-render the balance inside the hook in the RainbowKit usage. We should test to confirm this (if you haven't already). That would reduce the need for us to have a separate function for them to call, as it may already refetch based on their dapp usage (or they could force it that way)
- We also already have the `useAddRecentTransaction` hook and the `TransactionStoreProvider` which allows a dev to tell RainbowKit about a transaction to follow. This API is due for a refactor, but we generally want devs to funnel their transactions into that store so that RainbowKit can react to them. That reactive functionality should suffice for these refetch balance needs. We already poll to follow the status of a transaction there, so we could simply call `useBalance` again internally upon each tx confirmation.

packages/rainbowkit/src/transactions/TransactionStoreContext.tsx
```diff
@@ -39,6 +56,12 @@ export function TransactionStoreProvider({
     }
   }, [store, address, chainId]);
 
+  useEffect(() => {
+    if (store && address && chainId) {
+      return store.onTransactionStatus(onTransactionStatus);
+    }
+  }, [store, address, chainId, onTransactionStatus]);
```
magiziz, line 63: We could pass this logic in [here](https://github.com/rainbow-me/rainbowkit/blob/main/packages/rainbowkit/src/transactions/useRecentTransactions.ts#L20-L22) which would be less work, but there are some issues about doing that.

-Since `useBalance` from wagmi doesn't cause a automatic re-render when a transaction is confirmed. This means we'll have to do something on our end to cause a re-render and if we do it with [notifyListeners](https://github.com/rainbow-me/rainbowkit/blob/d5c3f2415b1e98b83b0f533e64e235ed6d980074/packages/rainbowkit/src/transactions/transactionStore.ts#L196) it would cause a lot of re-render to our `useRealtimeBalance` hook.
- We'll separate this logic to use our on `onTransactionStatus` event which will only pass us a status for a transaction when it's needed and when it's neccesarry. This way we'll cause a re-render to `useRealtimeBalance` and it won't waste too much RPC calls either.

magiziz, line 63: We could pass this logic in [here](https://github.com/rainbow-me/rainbowkit/blob/main/packages/rainbowkit/src/transactions/useRecentTransactions.ts#L20-L22) which would be less work, but that would cause a lot of re-renders + make unnecessary RPC calls since it'll happen pretty often due to how our transaction store provider works. 

We'll call refetch whenever is necessary to update our balance (e.g when a tx has successfully passed).

packages/rainbowkit/src/transactions/TransactionStoreContext.tsx
```diff
@@ -20,13 +26,26 @@ export function TransactionStoreProvider({
   const provider = usePublicClient() as PublicClient;
   const { address } = useAccount();
   const chainId = useChainId();
+  const { refetch } = useBalance({
+    address,
+    query: {
+      enabled: false,
+    },
+  });
```
DanielSinclair, line 34: @magiziz Should we move this logic to `useProfile` so that we're caching it and reduce the number of balance fetches? Currently I believe this hook would be called even if showBalance is disabled, and would cause duplicate balance fetch calls when rendering Profile -> Transaction UI

DanielSinclair (approved):
Everything else looks good; just need that 1 balance fetch optimization I think

DanielSinclair (approved):

PR #1722, opened 2024-01-23
fix: only call `WebkitTouchCallout` if `isIOS`

0xmkh:


KosmosKey:
Closing since we already have `touchCallout` in place for ios only

PR #1736, opened 2024-01-27
chore: deprecate `i18n-js` package with custom `i18n` class translation

0xmkh:
## Changes
- Decreased bundle size
- Removed `i18n-js`
- Created custom `I18n` translation class which should work exactly the same as `i18n-js`. We have `t` for translations and `onChange` to re-render components whenever `locale` gets updated

## What to test
1. Make sure the translations works the same as our [current example app](https://rainbowkit-example.vercel.app/)
2. Change locales and see if the translations works as expected. Also disable the `locale` prop and see if browser locale detection works as well
3. Optional: feel free to look at the unit test file [here](https://github.com/rainbow-me/rainbowkit/blob/9b9755c11148a61c8943b0dda4daa42d1d588cf8/packages/rainbowkit/src/classes/I18n.test.ts)

_Note: The main difference between `i18n-js` and the custom `I18n` class is that `i18n-js` does a lot of heavy checks and computations. On the other hand the custom `I18n` class gets translations through direct key access and the translations are [flattened](https://github.com/rainbow-me/rainbowkit/blob/%40mago/i18n-deprecation/packages/rainbowkit/src/classes/I18n.ts#L37-L60)._


akshatmittal:
You are on a roll @KosmosKey! This was exactly what I was discussing with a buddy last week and was going to create an issue with this _exact_ recommendation!

DanielSinclair (approved):

PR #1745, opened 2024-02-04
chore: upgrade wagmi, @tanstack/react-query, viem

0xmkh:


packages/create-rainbowkit/generated-test-app/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "generated-test-app",
   "private": true,
-  "version": "0.1.78",
+  "version": "0.1.0",
```
DanielSinclair, line 4: @KosmosKey Small issue here

PR #1760, opened 2024-02-08
fix: `<ConnectButton.Custom />` not displaying balance

### LINKED ISSUE #1750: [bug] Wallet balance not working

### RainbowKit Version

1.3.4

### wagmi Version

1.4.13

### Current Behavior

account.displayBalance is not working anymore and doesn't show anything

### Expected Behavior

account.displayBalance is supposed to show wallet balance

### Steps To Reproduce

Connect wallet and check for balance

_No response_

### Anything else?

Works fine to 1.3.2

0xmkh:
## Changes
- `<ConnectButton.Custom />` doesn't give balance back due to our `showBalance` context API blocking it. It's now forced to show balance. 
- With `<ConnectButton />` we wait until context API gets updated and then decide whether we show the balance or not. This way we don't show balance right away, but rather wait until we get some value from the context API.

## What to test
1. Do not test `<ConnectButton />` and `<ConnectButton.Custom />` at the same time since it'll force call rpc balance if you use the custom one.
2. Make sure when testing `<ConnectButton.Custom />` it should show balance no matter what. Look at network tab and see that the rpc is being called.
3. Make sure when testing with `<ConnectButton />` it should work as is right now with different responsive screen sizes.

_We don't want to expose another prop to our `<ConnectButton.Custom />` so we'll try to use context API as much as possible here 👍_


packages/rainbowkit/src/components/ConnectButton/ConnectButton.tsx
```diff
@@ -42,248 +42,265 @@ export function ConnectButton({
...
-        account,
-        chain,
-        mounted,
-        openAccountModal,
-        openChainModal,
-        openConnectModal,
-      }) => {
-        const ready = mounted && connectionStatus !== 'loading';
-        const unsupportedChain = chain?.unsupported ?? false;
+    ready && (
```
KosmosKey, line 56: Only thing i did here was `ready && ...` didn't touch any other elements of the component

DanielSinclair (approved):

PR #1770, opened 2024-02-14
fix: `showBalance` boolean support with `eth_getBalance` RPC logic

### LINKED ISSUE #1769: [bug] showBalance is not working as it should

### RainbowKit Version

2.0.0

### wagmi Version

2.5.5

### Current Behavior

i was experiencing some issues with rainbow 1.4.x and i found out that a new version released , so i created a new project out of the box using npm init @rainbow-me/rainbowkit@latest , just added showBalance={true} to show the balance and still not working , i m on the ethereum's mainnet.

### Expected Behavior

the showBalance when set to true should show balance of any chain , when set to false shouldn't show the balance , i think that it's set to false by default

### Steps To Reproduce

npm init @rainbow-me/rainbowkit@latest
and add         <ConnectButton showBalance={true} />
Screenshot 2024-02-14 at 3 12 38 PM
Screenshot 2024-02-14 at 3 12 56 PM

_No response_

### Anything else?

_No response_

0xmkh:
## Changes
- Fixed a bug where `showBalance` prop on the `<ConnectButton />` didn't work right if you used a boolean instead of responsive object values.

## What to test
1. Make sure to not have two `<ConnectButton />` component at the same time when testing.
2. Make sure the `showBalance` boolean behaviour works as expected and make sure it either calls `eth_getBalance` in the network tab or not based on `true` or `false` boolean:
```
<ConnectButton
  showBalance={true} // or false
/>
```
3. Also make sure the responsive object values work as expected without any breaks:
`<ConnectButton showBalance={{ largeScreen: true, smallScreen: false }} />`

_Note: `createNormalizeValueFn` from `@vanilla-extract/sprinkles` doesn't return `largeScreen` responsive value if a boolean is specified._

### DIFF
--- a/.changeset/thin-jokes-enjoy.md
+++ b/.changeset/thin-jokes-enjoy.md
@@ -1,0 +1,5 @@
+---
+"@rainbow-me/rainbowkit": patch
+---
+
+Fixed a bug where `showBalance` prop on the `<ConnectButton />` didn't work right if you used a boolean instead of responsive object values. `createNormalizeValueFn` from `@vanilla-extract/sprinkles` doesn't return `largeScreen` responsive value if a boolean was specified.
--- a/.changeset/thin-jokes-enjoy.md
+++ b/.changeset/thin-jokes-enjoy.md
@@ -2,4 +2,4 @@
 "@rainbow-me/rainbowkit": patch
 ---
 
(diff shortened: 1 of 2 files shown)


greg-schrammel (approved):
👌

DanielSinclair (approved):

DanielSinclair:
Looks good to me ✅

PR #1793, opened 2024-02-22
feat: `showCurrentChain` prop for `<ConnectButton />`

### LINKED ISSUE #1418: [bug] Chains not shown for Connect Button if only one chain

### RainbowKit Version

0.12.8

### wagmi Version

0.12.8

### Current Behavior

The chain button that toggles switch chains dropdown would not show if there is only one chain available.

https://github.com/rainbow-me/rainbowkit/blob/main/packages/rainbowkit/src/components/ConnectButton/ConnectButton.tsx#L71C15-L71C15

### Expected Behavior

It should be a configurable option to hide the button if there is only one chain available.

Otherwise we'd need to always rely on custom connect button to show the current chain.

### Steps To Reproduce

_No response_

_No response_

### Anything else?

_No response_

0xmkh:
## Changes
- Added `showCurrentChain` prop to always show the current chain without having to open the chain modal

## What to test
1. Pass in `showCurrentChain` prop to `<ConnectButton />` by using responsive values like `largeScreen`, `smallScreen` and even just a plain boolean. Make sure it works as expected in the example app


packages/example/pages/index.tsx
```diff
@@ -436,6 +443,35 @@ const Example = ({ authEnabled }: AppContextProps) => {
...
+                <tr>
+                  <td>
+                    <label htmlFor="showBalance">showCurrentChain</label>
+                  </td>
+                  <td>
+                    <input
+                      checked={showCurrentChainLargeScreen}
+                      id="showCurrentChain"
+                      onChange={(event) => {
+                        setShowCurrentChainLargeScreen(
```
0xartcro, line 455: Looks like it's hooked the other way around. The first one should be for small screen?

KosmosKey: Good catch 👍. Thanks 🙏

DanielSinclair:
I'm not quite sold on this one. When experimenting, if `showCurrentChain` is active when multiple chains are available, you lose the ability to switch chains. This prop also clashes with `chainStatus` and it isn't immediately clear to someone reading the types or docs why they're different.

My alternative proposal:
- Change the behavior so that the chain button (without hover, click reaction or drop-down) is visible after connection even when just 1 chain is available
- Users can override this new default with `none` on `chainStatus` reactive values (though I don't think many would)
- I thought about maybe adding a `chainVisibility` prop and type to replace the `none` in `chainStatus`, and to support the usecase with the chain indicator appears even before the user connects, but that still feels like an unusual edge case for most devs

magiziz:
Closing this PR in favor of https://github.com/rainbow-me/rainbowkit/pull/1806

PR #1798, opened 2024-02-22
chore: use .ts instead of .tsx for useChainId hook

Imti:
Very small change, noticed when going through the code and thought I'd change it after I kept seeing it


KosmosKey:
@Imti Thanks!

DanielSinclair (approved):

PR #1800, opened 2024-02-24
fix: use relative links

enjoyooor:
Using relative links as per [this wiki](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#relative-links)

I believe it's cleaner just in case there's a change to the org/repo name, etc.


DanielSinclair (approved):

DanielSinclair:
Thanks!

PR #1803, opened 2024-02-26
chore: adjust example webpack configs

DanielSinclair:
- Deprecated node module fallbacks that were required by walletconnect v1 libraries
- Updated installation docs


0xmkh (approved):
Clean 👏

PR #1805, opened 2024-02-26
chore: upgrade remix example to v2

DanielSinclair:
- upgrade remix to v2
- migrated remix config
- updated remix installation docs
- bump node engines in examples to v18


0xmkh (approved):
Looks good

PR #1814, opened 2024-02-29
fix: enable ens fetching only if `mainnet` is configured

### LINKED ISSUE #1813: [bug] `useMainnetEnsName` no longer checks if `mainnet` is enabled

### RainbowKit Version

2.0.1

### wagmi Version

2.5.7

### Current Behavior

Getting hard throws when `mainnet` is not part of the configured chains when trying to load Account Modal.

### Expected Behavior

Should not throw if `mainnet` is not configured.

### Steps To Reproduce

- Start a project with latest RainbowKit
- Configure chains as `[base, zora]` and omit `[mainnet]`
- Connect a wallet

_No response_

### Anything else?

Looking at what might have changed recently on the package —

`useMainnetEnsName` is no longer checking whether or not `mainnet` is enabled as a configured chain. This specific check was introduced at https://github.com/rainbow-me/rainbowkit/commit/fd08aa1f9929269c04aad7636bf9c2417d7dfe78 to resolve issues when Ens name was being fetched for `AccountModal` even when `mainnet` is not configured. This was reverted with https://github.com/rainbow-me/rainbowkit/commit/aa0269e49655da2638fda0c6853a746d55020878. Resulting in hard crashes.

```
ChainNotConfiguredError: Chain not configured.
```

Not sure if this was done since Wagmi should be handling this case, or if there is additional config needed on our end to bypass, or it is expected to work as is and its a client implementation error.

0xmkh:
## Changes
- If `mainnet` is not configured and cache is cleared you might get thrown with `ChainNotConfiguredError`

### DIFF
--- a/.changeset/popular-ducks-travel.md
+++ b/.changeset/popular-ducks-travel.md
@@ -1,0 +1,5 @@
+---
+"@rainbow-me/rainbowkit": patch
+---
+
+Fixed a bug where wagmi would throw `ChainNotConfiguredError` if `mainnet` is not configured as a chain. This is happening when fetching ens name and ens avatar.
--- a/packages/rainbowkit/src/hooks/useIsMainnetConfigured.ts
+++ b/packages/rainbowkit/src/hooks/useIsMainnetConfigured.ts
@@ -1,0 +1,14 @@
+import { mainnet } from 'wagmi/chains';
+import { useRainbowKitChains } from '../components/RainbowKitProvider/RainbowKitChainContext';
+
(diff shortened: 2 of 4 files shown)


DanielSinclair (approved):
Looks good. Surprised Wagmi/Viem doesn't support cross-chain resolution yet. Can't find any reference in their docs. Would be nice if we could make that possible; otherwise may need to start using an off-chain resolver service so that ENS resolutions works as you'd expect everywhere.

DanielSinclair (approved):

PR #1817, opened 2024-03-03
chore: `injectedWallet` on mobile detection & cleanup connectors code

0xmkh:
## Changes
- `connectorsForWallets` now appends `injectedWallet` automatically on mobile if `window.ethereum` is detected

## What to test
1. Open up metamask on mobile (or any other mobile wallets you use that supports browser wallet)
2. Paste in [this](https://rainbowkit-example-790kz443k-rainbowdotme.vercel.app/) preview link into the mobile wallet's browser
3. Make sure you get the browser wallet in the connect modal
4. Click on the browser wallet and make sure you're connected
5. Make sure `Sign Message`, `Sign Typed Data` and `Send Transaction` works as expected in the demo app.

_[Here](https://github.com/rainbow-me/rainbowkit/discussions/1732) is a related discussion that someone wanted to extend EIP-6963 to support mobile, but in this case many mobile wallet don't expose `rdns` which will result into a duplication issue. This also happens in the official WalletConnect modal_ 

_In this case Browser wallet should work fine since there will be only one `window.ethereum` provider and can't be overidden on mobile_

### DIFF
--- a/.changeset/weak-bears-brush.md
+++ b/.changeset/weak-bears-brush.md
@@ -1,0 +1,60 @@
+---
+"@rainbow-me/rainbowkit": patch
+"site": patch
+---
+
+`injectedWallet` gets added on its own if `window.ethereum` is found on mobile and it's not already on the custom wallet list. This way, you don't have to use WalletConnect if you don't need to and can stick with the built-in provider `window.ethereum` instead.
+
+You can also set `injectedWalletOnMobile` to `false` if you prefer not to use this feature. This can be done in `getDefaultConfig` as the first parameter or in `connectorsForWallets` as the second parameter.
+
+Example with `getDefaultConfig`
+
+```tsx
+import { getDefaultConfig } from "@rainbow-me/rainbowkit";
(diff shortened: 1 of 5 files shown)


DanielSinclair:
@magiziz Looks like we have a bit of a regression here. The expected behavior in the MetaMask dApp browser (or other dApp browsers) would be that `injectedWallet` (aka Browser Wallet) only appears when MetaMask is not included in the wallet list. It's essentially a fallback option. If `window.ethereum` is available as `isMetaMask` (using our other provider ignore list) or rdns for EIP-6963, then the MetaMask button would function as you'd expect to connect (and we favor EIP-6963 over `window.ethereum` when available). If the wallet auto-connects on a first or future session (like a previous approval, then visits dApp again), you also shouldn't need to go through the RainbowKit connect prompt. The Rainbow dApp browser coming up will support EIP-6963 by default, so I think we want to make sure we get this resolved. Another way to address that user's feedback is that we probably should simply re-order the mobile wallet list and bump up the active walletconnector the first position (so in this case MetaMask with `injectedWallet` hidden).

PR #1822, opened 2024-03-04
feat: export useAuthenticationStatus

tinystone007:
I need to export useAuthenticationStatus so I can get the uthenticationStatus from anywhere


DanielSinclair:
@tinystone007 Typically you would want to rely on the upstream authentication library for this behavior, as the `useAuthenticationStatus` internal to RainbowKit is used for specific UI logic. Does the snippet below for `next-auth` solve your needs?

```tsx
import { useSession } from "next-auth/react"

export default function Component() {
  const { data: session, status } = useSession()

  if (status === "authenticated") {
    return <p>Signed in</p>
  }
}
```

PR #1850, opened 2024-03-12
chore: Transports guide - installation page (docs)

### LINKED ISSUE #1849: [docs] [offer] Add transfer rpc instructions as they are missing since providers were deprecated in viem/wagmi

### RainbowKit Version

x.x.x

### wagmi Version

2.x

### Current Behavior

**This is an offer to write up docs, please indicate if a PR is appreciated or not.**.

The docs lack information on how to add your RPC. Finding out how to add alchemy/infura/other rpcs requires diving into the viem docs, which are far less developer friendly than rainbowkit attempts to be.

### Expected Behavior

Basic docs on wagmi/viem compatible transports config. The quickstart should include info on adding RPCs as this is not a luxury step but a best practice for production apps.

0xmkh:
## Changes
- Updated docs to showcase how to use `transports` with `getDefaultConfig`

### DIFF
--- a/.changeset/new-badgers-jog.md
+++ b/.changeset/new-badgers-jog.md
@@ -1,0 +1,5 @@
+---
+"site": patch
+---
+
+Added a section for how to add custom RPC in chains page.
--- a/.changeset/new-badgers-jog.md
+++ b/.changeset/new-badgers-jog.md
@@ -2,4 +2,4 @@
 "site": patch
 ---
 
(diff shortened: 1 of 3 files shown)


DanielSinclair:
@magiziz I think we'd want to move this to `Installation` instead, maybe with a section under "Additional build tooling setup" like `Preparing to deploy`. I'd also use Wagmi's `Transports` language to be more clear, because Custom RPC typically refers to adding a network on the wallet. I think we would want to explain why devs should use a production node like Alchemy or Quicknode there.

DanielSinclair (approved):

PR #1866, opened 2024-03-16
fix: `normalize` throws an error for disallowed characters

### LINKED ISSUE #1865: [bug] invalid ENS name crashes application

### RainbowKit Version

2.0.1

### wagmi Version

2.5.7

### Current Behavior

connecting to an app with an invalid ENS name will crash an entire app

### Expected Behavior

connecting should just not display the invalid name

### Steps To Reproduce

1. visit any rk app (e.g. rainbowkit.com)
2. connect via an account with an invalid name (`0x2031734704dcc0c56f5588dc737851154b44651b` has one)
3. app will crash entirely

_No response_

### Anything else?

_No response_

0xmkh:
## Changes
- Added a new function called `safeNormalize` with a try/catch block


packages/rainbowkit/src/hooks/useMainnetEnsAvatar.ts
```diff
@@ -6,9 +6,17 @@ import { useIsMainnetConfigured } from './useIsMainnetConfigured';
...
+      return normalize(ensName);
+    } catch {
+      return ensName;
+    }
+  };
+
   const { data: ensAvatar } = useEnsAvatar({
     chainId: mainnet.id,
-    name: name ? normalize(name) : undefined,
+    name: name ? safeNormalize(name) : undefined,
```
DanielSinclair, line 19: I think this would still fail because the raw `ensName` returned wouldn't be normalized and then `useEnsAvatar` would fail in a cascade. I think we should either just get the normalization fixed upstream in viem or skip ENS resolution when `normalize` throws an error.
Screenshot 2024-03-19 at 2 02 55 AM

DanielSinclair (changes requested):

DanielSinclair (approved):
Looks good to me!

PR #1871, opened 2024-03-19
feat: optimistic disconnect

0xmkh:
## Changes
- Added optimistic disconnect behavior by checking `isDisconnecting` in `RainbowKitWagmiStateProvider`. If `isDisconnecting` is `true`, the `useConnectionStatus` hook is set to the `disconnected` state.

https://github.com/rainbow-me/rainbowkit/assets/53529533/38cbff5e-465a-41d2-b924-e613abca32fd


DanielSinclair:
@magiziz For that WalletConnect scenario, I think what might work better is an "optimistic disconnect".  When the user taps disconnect, we put the RainbowKit modal into the disconnected state and then trigger the `useDisconnect` hook internally. This way the user is never blocked; they don't need to return to their wallet to disconnect anyway. We may need a retry to ensure Wagmi's state is in sync if the websocket were to fail

site/data/en-US/docs/custom-connect-button.mdx
```diff
@@ -19,6 +19,7 @@ export const YourApp = () => {
     <ConnectButton.Custom>
       {({
         account,
+        disconnecting,
```
magiziz, line 22: This technically won't be a breaking change. If we were to merge this the only thing that would impact is if `disconnecting` is true, then `openConnectModal` with just return a `noop` function. [Here](https://github.com/rainbow-me/rainbowkit/blob/main/packages/rainbowkit/src/components/ConnectButton/ConnectButtonRenderer.tsx#L154-L156) is an example.

magiziz: For injected wallets it should happen super fast, but for WalletConnect in some scenarios might take time. Of course depending on their socket server.

DanielSinclair:
I started to look into a refactor here:
- I don't think `RainbowKitWagmiStateContext` is necessary as we have the existing `useConnectionStatus` hook
- `useDisconnect` also comes with a `status` prop that we could listen to globally and reflect in `useConnectionStatus`
- Throughout the codebase we use `isConnected` and `isConnecting` from `useAccount`; we would want to consolidate this to only use `useConnectionStatus`
- As long as the disconnect `status` flips to `pending` after a user was in a connected status, then we know they're in the disconnecting state
- We would just want to treat `disconnecting` like existing connection status `disconnected` throughout the uses of that hook.

packages/rainbowkit/src/components/RainbowKitProvider/RainbowKitWagmiStateProvider.tsx
magiziz, line 1: I know this is not ideal especially in the long term, but wagmi's `useDisconnect` hook doesn't provide a global scope state support. It only supports at a component level since they use TanStack react query mutation directly. 

Maybe we can look at having `isDisconnecting` state in `useAccount` hook in the future, but for now this approach works.

As long as we mark `isDisconnecting` to `disconnect` mode in `useConnectionStatus` that'd work for now 🙏

packages/rainbowkit/src/hooks/useConnectionStatus.ts
```diff
@@ -10,8 +11,9 @@ export type ConnectionStatus =
 export function useConnectionStatus(): ConnectionStatus {
   const authenticationStatus = useAuthenticationStatus();
   const { isConnected } = useAccount();
+  const { isDisconnecting } = useRainbowKitWagmiState();
 
-  if (!isConnected) {
+  if (!isConnected || isDisconnecting) {
     return 'disconnected';
   }
```
magiziz, line 19: This is the most important part out of all which makes this work

packages/rainbowkit/src/components/RainbowKitProvider/RainbowKitWagmiStateProvider.tsx
magiziz, line 1: I know this is not ideal especially in the long term, but wagmi's `useDisconnect` hook doesn't provide a global scope state support. It only supports at a component level since they use TanStack react query mutation directly. 

Maybe we can look at having `isDisconnecting` state in `useAccount` hook in the future, but for now this approach works.

As long as we mark `isDisconnecting` to `disconnected` mode in `useConnectionStatus` that'd work for now 🙏

packages/rainbowkit/src/components/RainbowKitProvider/ModalContext.tsx
```diff
@@ -119,13 +122,18 @@ export function ModalProvider({ children }: ModalProviderProps) {
...
           openConnectModal:
-            connectionStatus === 'disconnected' ||
-            connectionStatus === 'unauthenticated'
+            // Prevent opening the connect modal during disconnecting mode. Even though `connectionStatus`
+            // may mark it as 'disconnected', we don't want the user to open the modal as the wagmi state
+            // still considers it 'connected'
+            !isDisconnecting &&
+            (connectionStatus === 'disconnected' ||
+              connectionStatus === 'unauthenticated')
               ? openConnectModal
```
magiziz, line 131: You technically can still open the connect modal if your `connectionStatus` is returning `disconnected`. 

If we allow users to open the connect modal by the time WalletConnect takes time to disconnect, they'll face the issue with shim disconnection which is a known issue in wagmi rn. Ofc only if they start connecting different wallets at once, but we'll make sure to disable it.

PR #1890, opened 2024-03-26
fix: await connectWallet for walletbutton

jguyton:
What changed:
- updated WalletButtonRendererProps `connect` type to be a Promise matching the handler returned
- await  `connectWallet` promise during connect


0xmkh (approved):
🙏

PR #1893, opened 2024-03-27
fix: compatible wagmi config type for `getDefaultConfig`

ponjs:
Compatible with wagmi [config type](https://wagmi.sh/react/typescript#config-types)

```ts
const config = getDefaultConfig({
  appName: 'RainbowKit demo',
  projectId: 'YOUR_PROJECT_ID',
})

declare module 'wagmi' { 
  interface Register { 
    config: typeof config 
  } 
} 
```


ponjs:
> @ponjs Hello! What issue are you aiming to fix here? Can you add an explainer?

Because the current config type prompts `'config' is referenced directly or indirectly in its own type annotation`.

DanielSinclair:
@ponjs What versions of Wagmi and Viem are you currently using?

ponjs:
> @ponjs What versions of Wagmi and Viem are you currently using?

```
"@rainbow-me/rainbowkit": "^2.0.2",
"viem": "^2.9.3",
"wagmi": "^2.5.13",
```

magiziz:
@ponjs Just realised that `WagmiProviderProps['config']` is no longer being used as the type in their `<WagmiProvider />` component with `config` prop. Did some cleanup for you to not return `WagmiProviderProps['config']` as a return prop for `getDefaultConfig`.

ponjs:
> @ponjs Just noticed that `WagmiProviderProps['config']` is kind of funky with `createConfig` type itself. Did some cleanup for you to not return `WagmiProviderProps['config']` as a return prop for `getDefaultConfig`.

Hello, but this is missing the generic type, which prevents it from inferring the value of chainId, since declaring the type of config serves the purpose of needing to get the value of chainId. The [documentation](https://wagmi.sh/react/typescript#declaration-merging) also mentions.

image

magiziz:
@ponjs Just made type even more stricter. Should work now. Let me know if it looks good 👍

image

ponjs:
> @ponjs Just made type even more stricter. Should work now. Let me know if it looks good 👍
> 
> image

👍 Thanks, that's what I was trying to achieve

DanielSinclair (approved):

PR #1897, opened 2024-03-28
feat: add `zetachain` metadata

Azleal:
add zetachain icon


DanielSinclair:
@Azleal Do you happen to have a ZetaChain SVG with an opaque colored background?

Azleal:
> @Azleal Do you happen to have a ZetaChain SVG with an opaque colored background?

@DanielSinclair I updated the zetachain svg logo with an opaque colored background. also added zetachain to site project.

0xmkh (approved):
Looks good 👍

Did a small tweak and renamed testnet zetachain to `Zeta Athens` instead of `ZetaChain Athens` since it might cause overflow UI bug to the modal.

PR #1923, opened 2024-04-10
fix: let `metaMaskWallet` connect to rainbow wallet if overidden

0xmkh:
## Changes
- When metamask and rainbow wallet are injected, `metaMaskWallet` ignores rainbow wallet in `window.ethereum.providers`. This means if a user is not using EIP-6963 feature, then rainbow wallet won't work if someone clicks on metamask.

### DIFF
--- a/.changeset/fast-files-clean.md
+++ b/.changeset/fast-files-clean.md
@@ -2,4 +2,4 @@
 "@rainbow-me/rainbowkit": patch
 ---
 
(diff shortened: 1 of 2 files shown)


DanielSinclair:
The MetaMask button seems like it working as expected without this change.

DanielSinclair:
Do you have a path to reproduce the issue?

magiziz:
Closing this since we already have `window.ethereum` as a fallback

PR #1924, opened 2024-04-10
fix: metamask button fallback (`1.3.x`)

0xmkh:
## Changes
- When metamask and rainbow wallet are injected, `metaMaskWallet` ignores rainbow wallet in `window.ethereum.providers`. This means if a user is trying to connect to metamask when they have rainbow wallet it won't connect to rainbow wallet.


DanielSinclair:
Discovered that this actually breaks the expected behavior when both Rainbow and MetaMask are injected, where we want the MetaMask button to always point to MetaMask.

PR #1806, opened 2024-02-26
feat: chain modal v2

0xmkh:
## Changes
- Now you can choose any chains even if your wallet is not connected.
- Works with `siwe`.
- During connection process, it automatically selects the chain you picked earlier. Or the one you set as your default `initialChain` instead.
- If you've set `initialChain`, the option to change networks will be hidden to prevent accidental switches.
- The current chain will show up only even if you have one chain unless you set `chainStatus` to `none`

## Screen recordings / screenshots

https://github.com/rainbow-me/rainbowkit/assets/53529533/fab003eb-5526-45ca-976d-f8a8afc2e702

## What to test
1. Checkbox the `chainModalOnConnect` in the example app
2. Try opening the chain modal and switch through your desired chain
3. While connecting your wallet, confirm it connects to the chain you chose.
4. Set `initialChain` and check if the chain button is no longer visible.
5. Make sure everything works well with `siwe`. Test this connecting to your wallet first and then you should see chain button dissapear. This prevents users from switching chains with their connected wallet when they are not signed in yet.

PR #1927, opened 2024-04-15
feat: pass openChainModal even if connectionStatus is not connected

hungdoansy:
Some DEXes often support multiple chains. Users tend to play around without connecting to the websites.

With the current value of ModalContext, it's not possible for users to switch chains without connecting wallets.

### DIFF
--- a/packages/rainbowkit/src/components/RainbowKitProvider/ModalContext.tsx
+++ b/packages/rainbowkit/src/components/RainbowKitProvider/ModalContext.tsx
@@ -116,8 +116,7 @@ export function ModalProvider({ children }: ModalProviderProps) {
             isCurrentChainSupported && connectionStatus === 'connected'
               ? openAccountModal
               : undefined,
-          openChainModal:
-            connectionStatus === 'connected' ? openChainModal : undefined,
+          openChainModal,
           openConnectModal:
             connectionStatus === 'disconnected' ||
             connectionStatus === 'unauthenticated'


hungdoansy:
Hi @magiziz, can you please review this PR?

hungdoansy:
> @hungdoansy Can you explain more why this change is needed ? Ideally we would want to open the chain modal after user has authenticated / connected their wallet, but lmk your what you think.

I believe that people should be able to change the network to play with the webapp even tho they haven't connected yet.
Why?
Some people don't have wallet or they just don't want to connect to strange websites. We shouldn't prevent them playing around with the webapp. 

How'd you think?

0xartcro:
> > @hungdoansy Can you explain more why this change is needed ? Ideally we would want to open the chain modal after user has authenticated / connected their wallet, but lmk your what you think.
> 
> I believe that people should be able to change the network to play with the webapp even tho they haven't connected yet. Why? Some people don't have wallet or they just don't want to connect to strange websites. We shouldn't prevent them playing around with the webapp.
> 
> How'd you think?

https://github.com/rainbow-me/rainbowkit/pull/1806 Should fix that I believe

magiziz:
> > @hungdoansy Can you explain more why this change is needed ? Ideally we would want to open the chain modal after user has authenticated / connected their wallet, but lmk your what you think.
> 
> I believe that people should be able to change the network to play with the webapp even tho they haven't connected yet. Why? Some people don't have wallet or they just don't want to connect to strange websites. We shouldn't prevent them playing around with the webapp.
> 
> How'd you think?

Right. That change will also cause users to switch chains without having their wallet connected which would result to unexpected behavior and like @0xartcro mentioned the PR https://github.com/rainbow-me/rainbowkit/pull/1806 should fix this behaviour.

hungdoansy:
Thanks @magiziz @0xartcro for letting me know. Just took a look at the preview link. It's exactly what I needed. I'll close this PR now.

PR #1934, opened 2024-04-17
feat: enhanced provider API / `@tanstack/react-query` peer dependency

0xmkh:
## Changes
- Added `@tanstack/react-query` as a peer dependency and introduced enhanced provider API.
- Implemented fallback to fetch ENS name from enhanced provider API when mainnet is unavailable.

## What to test
- Go to this preview link: https://rainbowkit-example-h3v2divmt-rainbowdotme.vercel.app/
- Open up network tab by right clicking -> inspect element -> network tab
- If you connect to a wallet that has ENS name you should be able to see it making request to our enhanced provider API


DanielSinclair:
@magiziz Noticed the address -> resolved name is a bit slow, even after the first fetch. Seeing a few network requests after leaving the page idle too. Are we taking advantage of react-query caching? https://github.com/rainbow-me/rainbowkit/assets/4412473/307f133e-0b26-49dd-b092-91b215adaf9d

DanielSinclair:
Also pushed up some tweaks, mainly one that injects the API key at build-time. This can be a bit messy because it also needs to be available within the CI tests (where we can't statically replace the API key). Lmk if you have other thoughts! It will be skipped in CI here because we need to add a new ENV.

DanielSinclair (changes requested):
When replicating a 401, I noticed we throw console errors and try to re-fetch every few seconds. We should fail more gracefully because we may wind down this API endpoint some day. Max retry should be limited, and we should only need to fetch once at first-render if the caching is working appropriately 
Screenshot 2024-05-07 at 2 36 14 AM

packages/rainbowkit/src/core/network/enhancedProvider.test.ts
```diff
@@ -0,0 +1,19 @@
+import { describe, expect, it } from 'vitest';
+import { enhancedProviderHttp } from './enhancedProvider';
+
+describe.skip('createHttpClient', () => {
```
magiziz, line 4: Will do a follow up PR to fix this unit test. Faces issues with github secrets and `process.env`, but we can run this locally for now.

DanielSinclair:
Force pushed to cleanup the git history here

DanielSinclair (approved):

PR #1935, opened 2024-04-17
chore: `coinbaseWalletParameters` for `connectorsForWallets`

0xmkh:
## Changes
- Some users right now can't pass in `coinbaseWallet` parameters directly in `connectorsForWallets`. This PR fixes it.

### DIFF
--- a/.changeset/curvy-parents-push.md
+++ b/.changeset/curvy-parents-push.md
@@ -1,0 +1,5 @@
+---
+"@rainbow-me/rainbowkit": patch
+---
+
+Introduced `coinbaseWalletParameters` parameter in `connectorsForWallets` to support `coinbaseWallet` connector parameters.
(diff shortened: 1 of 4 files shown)


DanielSinclair:
v4 deprecates the following options, which is almost all of them, so I think we'll need to close this and re-evaluate any options we need to expose once our v4 upgrade lands:
```
enableMobileWalletLink (enabled by default in v4)
jsonRpcUrl
reloadOnDisconnect
uiConstructor
overrideIsMetaMask
overrideIsCoinbaseWallet
diagnosticLogger
reloadOnDisconnect
headlessMode
```
https://github.com/coinbase/coinbase-wallet-sdk/blob/master/docs/migration_guide.md

DanielSinclair:
Will leave the branch hanging in case we want to revist/reopen

PR #1953, opened 2024-04-30
fix: coinbase wallet redirect behavior

DanielSinclair:
## Changes
- Mitigated behavior in Coinbase Wallet where the user would be captured in the in-app browser upon redirect.
- Improved QR Code error correction reliability to assist wallets that sometimes fail to scan a RainbowKit QR Code.


0xmkh (approved):
Works great

PR #1964, opened 2024-05-07
feat: coinbase smart wallet support

0xmkh:


packages/rainbowkit/src/types/utils.ts
```diff
@@ -11,3 +13,77 @@ export type Omit<type, keys extends keyof type> = Pick<
...
+      | undefined;
+  }
+>;
+
+export type WindowProvider = {
+  coinbaseWalletExtension?: WalletProvider | undefined;
+  ethereum?: WalletProvider | undefined;
+  phantom?: { ethereum: WalletProvider } | undefined;
+  providers?: any[] | undefined; // Adjust the type as needed
+};
```
magiziz, line 89: Here is example from wagmi https://github.com/wevm/wagmi/blob/main/packages/core/src/connectors/injected.ts#L636-L640

magiziz, line 89: [Here](https://github.com/wevm/wagmi/blob/main/packages/core/src/connectors/injected.ts#L636-L640) is a example from wagmi

packages/rainbowkit/src/window.d.ts
magiziz, line 1: `wagmi/window` is deprecated

packages/rainbowkit/src/connectors/coinbaseWallet.ts
jxom, line 1: Should be able to use `coinbaseWallet` in Wagmi now.

DanielSinclair (approved):

PR #2004, opened 2024-05-28
feat: rainbowkitprovider add classname

sj719045032:
When I used the theme attribute, I could not customize the style of the outer div, so I added the className attribute so that I could customize the style of the div, such as the height, etc


magiziz:
We're going to close this PR for now since we'll working on another fix for this.

sj719045032:
> @sj719045032 Are you looking to add height and width only ? We usually prefer using theme instead of className. Can you describe exactly what you're looking for ? I'll look and see what we could do on our end.

yes，So how can I easily set the height?

DanielSinclair:
@sj719045032 Are you referring to the modal height, or the wrapper height?

PR #2005, opened 2024-05-28
chore: refactor wagmi provider in examples

DanielSinclair:
## Changes
- migrated wagmi config to `src/` dir for next and vite examples, added wagmi config file


0xmkh (approved):
lgtm

PR #2017, opened 2024-05-30
feat: add `preference` property to `coinbaseWallet`

0xmkh:
## Changes
- Added `preference` property to `coinbaseWallet`


packages/rainbowkit/src/wallets/walletConnectors/coinbaseWallet/coinbaseWallet.ts
```diff
@@ -106,3 +107,10 @@ export const coinbaseWallet = ({
     },
   };
 };
+
+export const coinbaseWallet = Object.assign<
+  typeof wallet,
+  { preference: CoinbaseWalletParameters<'4'>['preference'] }
```
magiziz, line 113: This would give us cleaner typescript builds.

Here is how `coinbaseWallet.d.ts` build file would look like:

```
import { CoinbaseWalletParameters } from 'wagmi/connectors';
import { Wallet } from '../../Wallet';
export interface CoinbaseWalletOptions {
    appName: string;
    appIcon?: string;
}
export declare const coinbaseWallet: (({ appName, appIcon }: CoinbaseWalletOptions) => Wallet) & {
    preference: CoinbaseWalletParameters<'4'>['preference'];
};
```

DanielSinclair: Was there a reason we shifted away from the Interface type? I lean in that direction because the `preference` would be undefined unless the user specifies.

packages/example/pages/_app.tsx
```diff
@@ -146,6 +147,9 @@ const sei = {
   contracts: {},
 } as const satisfies Chain;
 
+// Enable coinbase smart wallet feature
+coinbaseWallet.preference = 'smartWalletOnly';
+
```
DanielSinclair, line 152: We'll want to enable the coinbase wallet option in the example dApp; we could test the types here by passing `all` though

magiziz: Yep. Done 👍

DanielSinclair (approved):

Issue #2036, opened 2024-06-07
[bug] <Cannot read properties of undefined (reading 'toLowerCase')> Navigator undefined in getBrowser


0xAskar:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### RainbowKit Version

2.1.2

### wagmi Version

2.9.10

### Current Behavior

The problem is that after migrating and following the migration guides to the 2.x rainbow wallet, and the same for wagmi and viem, I'm getting a weird problem where the navigator in the ```node_modules/@rainbow-me/rainbowkit/dist/index.js``` is not being recognized, such that the function, `getBrowser()` has a problem when `navigator.userAgent.toLowercase()` is called because navigator is an empty Object (I console.log'ed it and it showed that).

If I just return normally with an if statement checking if its empty, my code runs as functioned. But I assume you shouldn't be changing the node modules for rainbow kits. So any idea why this is happening? 

So, please see the full error below:
<img width="1081" alt="image" src="https://github.com/rainbow-me/rainbowkit/assets/37524379/650d2708-0507-41d8-bd1c-9b4e977b000b">

```
function getBrowser() {
  if (typeof navigator === "undefined")
    return "Browser" /* Browser */;
  const ua2 = navigator.userAgent.toLowerCase();
  ...
  ```

I've attached my app.js below
```
import React from "react";
import { GlobalPriceProvider } from "../context/GlobalPriceContext";
import "../styles/globals.css";
import Router from 'next/router';
import { NextUIProvider } from "@nextui-org/react";
import dotenv from "dotenv";
import NProgress from 'nprogress';
import 'nprogress/nprogress.css';
import "@rainbow-me/rainbowkit/styles.css";
import { ToastContainer, toast } from 'react-toastify';
import 'react-toastify/dist/ReactToastify.css';
import { SpeedInsights } from '@vercel/speed-insights/next';
import { Analytics } from "@vercel/analytics/react";
import {
  getDefaultConfig,
  RainbowKitProvider,
  darkTheme
} from "@rainbow-me/rainbowkit";
import { WagmiProvider, http } from 'wagmi';
import { CompModalProvider } from "context/CompModalContext";
import {
  RainbowKitSiweNextAuthProvider,
} from "@rainbow-me/rainbowkit-siwe-next-auth";
import { SessionProvider } from "next-auth/react";
import { mainnet, polygon, optimism, arbitrum, base } from "wagmi/chains";
[...]

PR #2039, opened 2024-06-09
fix: navigator object is empty

0xAskar:
Navigator object is empty in some cases and as a result throws an error in getBrowser preventing the user from using the app in dev mode. This checks if is an empty object and then returns Browser.browser, simple fix. reference to https://github.com/rainbow-me/rainbowkit/issues/2036#issuecomment-2156252957


0xAskar (commented):
any reason why you not keep the if statement and keep the optional chaining as well?

DanielSinclair (approved):

PR #2057, opened 2024-06-18
fix: reduce klaytn logo size

nezouse:
For some reason it was included in vite build even tho we don't use it. So at least let's make it small.


DanielSinclair (approved):

DanielSinclair:
Thanks! These SVGs should be handled by tree splitting, but maybe Vite handles this differently than Next.js

PR #2060, opened 2024-06-19
fix: add `safeWallet` to `getDefaultWallets`

0xmkh:
## Changes
- Adding back `safeWallet` and `braveWallet` to `getDefaultWallets`, since it got removed from v2. [Here](https://github.com/rainbow-me/rainbowkit/blob/1.x/packages/rainbowkit/src/wallets/getDefaultWallets.ts) is an example from v1.


DanielSinclair:
Tested with the Safe Browser by adding a Custom App here and not seeing the Safe connector appear; we might have a larger issue. Previously this connector would auto-connect in the Safe browser environment. https://app.safe.global/apps 
Screenshot 2024-06-27 at 1 33 38 AM

DanielSinclair (changes requested):
Issue above

magiziz:
@DanielSinclair Ah i see. I'll deeper debugging👍

Btw how were you able to get the RainbowKit example app working on on safe browser ? For me it shows empty / blank screen:

image

image

It happens with all other apps too.

magiziz:
@DanielSinclair Safe wallet seems to work fine now. Can you check ?

image

image

DanielSinclair (approved):

PR #2072, opened 2024-07-02
feat: `binanceWallet`

0xmkh:
## Changes
- Added `binanceWallet`


magiziz:
I used "Binance Wallet" as the wallet name instead of "Binance Web3 Wallet" since it kind of broke the connect modal because of it's long name.

DanielSinclair (approved):

PR #2078, opened 2024-07-03
fix: `metaMaskWallet` causing the app to crash due to `window.ethereum.providers` override

0xmkh:
### Changes
- Fixed an issue where `metaMaskWallet` would cause the app to crash if MetaMask, Coinbase, and Rainbow wallets were installed. This happened due to a conflict with `window.ethereum.providers`, which caused an infinite loop when these wallets overrode each other.


magiziz:
Closing this PR since it's a provider issue rather than RainbowKit issue

PR #2126, opened 2024-08-01
chore: tree shake chain and account modals

### LINKED ISSUE #2110: [bug] Tree-shaking isn't taking place for unused features

### RainbowKit Version

2.1.2

### wagmi Version

2.11.2

### Current Behavior

In cases where `AccountModal` or `ChainModal` aren't used, their code is still fully bundled by webpack.

The worst "offender" is the `normalize` function viem reexports (at 25kb gzipped), which is (probably?) imported `RainbowKitProvider -> ModalProvider -> AccountModal -> useProfile -> useMainnetEnsAvatar-> safeNormalize`.

### Expected Behavior

Rainbowkit should allow for the configuration of used/unused features, and structuring the code in such a way that disabled features are tree-shaked away.

### Steps To Reproduce

_No response_

_No response_

### Anything else?

_No response_

0xmkh:
## Changes

Applied tree shaking for chain and account modals to prevent them from bundling when unused.

Here you can see that account and chain modals are bundled separately if they're being used.

https://github.com/user-attachments/assets/8fac1595-89a8-4f01-8e63-7f83b406897a

## What to test

- Whenever you don't use `useAccountModal` hook you shouldn't see account modal being bundled in network tab.
- Whenever you don't use `useChainModal` hook you shouldn't see chain modal being bundled in network tab.

### DIFF
--- a/.changeset/wise-phones-refuse.md
+++ b/.changeset/wise-phones-refuse.md
@@ -1,0 +1,5 @@
+---
+"@rainbow-me/rainbowkit": patch
+---
+
+Implemented tree shaking for chain and account modals to prevent them from bundling when unused.
--- a/packages/rainbowkit/src/components/RainbowKitProvider/ModalContext.tsx
+++ b/packages/rainbowkit/src/components/RainbowKitProvider/ModalContext.tsx
@@ -1,6 +1,9 @@
 import React, {
+  Fragment,
   ReactNode,
+  Suspense,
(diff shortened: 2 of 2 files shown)


LufyCZ:
lgtm! Don't think there's a better solution. Thanks

LufyCZ:
@DanielSinclair any specific reason this was closed?

PR #2176, opened 2024-09-11
Refactor ModalProvider to remove wrapper div and apply theme props directly

### LINKED ISSUE #2164: [bug] Provider affects the DOM,

### RainbowKit Version

2.1.5

### wagmi Version

2.17.0

### Current Behavior

The `RainbowKitProvider` component injects a `<div>` which affects styling.

image

The above should result in a fully red screen, however, the "injected" div doesn't have a `h-full`, and thus the div below won't stretch to the height of the screen.

image

### Expected Behavior

The provider should not affect the DOM in a way that potentially breaks styling.

### Steps To Reproduce

_No response_

https://github.com/LufyCZ/rainbowkit-bug

### Anything else?

_No response_

Adebesin-Cell:
# Refactor ModalProvider to remove wrapper div and apply theme props directly

This PR refactors the ModalProvider component to remove the unnecessary wrapper div and apply theme properties directly to its children. This change allows for more flexible use of the ModalProvider throughout the component tree.

Key changes:
- Removed the wrapper div from ModalProvider
- Used React.Fragment to avoid adding extra DOM elements
- Applied theme properties directly to child elements using React.Children.map
- Conditionally rendered the style element only when a theme is provided

This refactor improves the flexibility of the ModalProvider, allowing it to be used at any point in the component tree without introducing additional DOM nesting. It maintains all existing theming functionality while providing a cleaner, more efficient implementation.

Testing:
- Verified that the ModalProvider works correctly with and without a theme
- Ensured that theme properties are correctly applied to child components
- Checked that the component renders correctly in different contexts without introducing unexpected DOM elements

Please review and test thoroughly to ensure this change doesn't introduce any unintended side effects in existing implementations.

Fixes https://github.com/rainbow-me/rainbowkit/issues/2164


DanielSinclair:
At this time, we don't need any refactors. If you're interested in contributing, feel free to investigate our open Issues and submit fixes, if necessary.

Adebesin-Cell:
> At this time, we don't need any refactors. If you're interested in contributing, feel free to investigate our open Issues and submit fixes, if necessary.

I think it was part of opened issues about DOM injection. 

cc: https://github.com/rainbow-me/rainbowkit/issues/2164

PR #2185, opened 2024-09-19
separate wagmi config params from config

frolic:
I want to use a custom Viem client with RainbowKit so that I can extend with a few custom actions, but the way RainbowKit constructs a Wagmi client means I can't make these changes on the client itself.

By separating out the `getDefaultConfigParameters` from `getDefaultConfig`, I can still lean on RainbowKit defaults, make adjustments to these params, then call `createConfig` with the adjusted params. This felt like an easier lift than adding support for the `client` arg here.

This change is backwards compatible, just relocates the param creation to its own file/export.


DanielSinclair:
@holic You would actually just want to use `getDefaultWallets` or `connectorsForWallets` API to get connectors configured by RainbowKit, and then directly use `createConfig`

PR #2251, opened 2024-11-26
feat: Flow chain support

nialexsan:
- add EVM on Flow to chains: https://developers.flow.com/evm/using


nialexsan:
https://github.com/onflow/docs/issues/1005

PR #2297, opened 2025-01-22
feat: use InjectedConnector in binance dapp browser

mame-col:
# Changes

In the original Binance Wallet configuration, `WalletConnectConnector` was used for connections in **all** cases. However, within the built-in browser of Binance Wallet, we provide an `InjectedProvider`, so using `InjectedConnector` is a better approach. This is because our Android environment currently does not support using WalletConnect within the built-in browser of Binance Wallet, which has caused some users to be unable to connect their wallets in dApps.

## Bad case video

https://github.com/user-attachments/assets/c8463026-9c8c-414a-b1c5-74cb0bbbffdb

## Modified behaviour video

https://github.com/user-attachments/assets/0ce07c5d-a757-4c8b-8e9d-91ef1c75605f

### DIFF
--- a/.changeset/breezy-tigers-work.md
+++ b/.changeset/breezy-tigers-work.md
@@ -1,0 +1,5 @@
+---
+"@rainbow-me/rainbowkit": patch
+---
+
+Improved support for the Binance Wallet dApp browser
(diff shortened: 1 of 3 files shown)


xuzhanhh:
Sorry, I accidentally clicked apply as code owner, please ignore it.

DanielSinclair (approved):

DanielSinclair:
Looks good to me! Thank you for the concise explanation and clean fix 🫡

PR #2304, opened 2025-01-31
feat: beam wallet support

royscheepens:
This PR adds the Beam SDK connector for apps that connect to the Beam SDK via RainbowKit. For more information see [https://github.com/BuildOnBeam/beam-sdk-web](https://github.com/BuildOnBeam/beam-sdk-web) and [https://docs.onbeam.com](https://docs.onbeam.com).

### DIFF
--- a/packages/example/src/wagmi.ts
+++ b/packages/example/src/wagmi.ts
@@ -1,6 +1,7 @@
 import { type Chain, getDefaultConfig } from '@rainbow-me/rainbowkit';
 import {
   argentWallet,
+  beamWallet,
   berasigWallet,
   bestWallet,
   bifrostWallet,
@@ -208,6 +209,7 @@ export const config = getDefaultConfig({
       groupName: 'Other',
       wallets: [
         argentWallet,
+        beamWallet,
         berasigWallet,
         bestWallet,
         bifrostWallet,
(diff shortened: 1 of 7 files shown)


DanielSinclair:
@royscheepens We are hesitant about adding additional SDKs. Can you explain how the SDK and Beam works, and what the value of this would be? Teams can always distribute a RainbowKit wallet connector externally that is compatible with the wallet list in RainbowKit.

RobbyUitbeijerse:
Hey @DanielSinclair , I'm in the same team as Roy :-)

Very fair take, and we don't think we're adding value to RainbowKit as much as we're trying to make our own service more credible by offering sensible defaults in libraries that might already be used by parties that (want to) work with us. We do the same for eg viem - [link](https://github.com/wevm/viem/blob/main/src/chains/index.ts#L46-L47), and ethereum-lists/chains - [link](https://github.com/ethereum-lists/chains/blob/master/_data/chains/eip155-13337.json)

We're already offering an connector externally through [@onbeam/utils](https://www.npmjs.com/package/@onbeam/utils), so we're mostly covered. Is there any chance RainbowKit is open to a new page on the rainbowkit.com docs where external adapters can be named and promoted? I guess that more would achieve the same thing, but we understand if that might not be an option too.

DanielSinclair:
@RobbyUitbeijerse Feel free to open a PR to add a `/guide` post, and we can add that to the docs site. Reference the directory [here]( https://github.com/rainbow-me/rainbowkit/tree/182440a9a9bda4bf5f721425472a18f434a63e4f/site/data/en-US/guides). I'll follow-up and add an index to link to these.

DanielSinclair:
Closing this wallet connector PR for now

PR #2305, opened 2025-01-31
fix: fix incorrect import of esbuild-plugin-replace in build.js

0xwitty:
The import for `esbuild-plugin-replace` was incorrect—it doesn't provide a named export `replace`.
This caused an error when building. Updated it to use the default import instead.


DanielSinclair:
Not sure that is true. What error are you seeing? Can you share info about your Node version and environment?

Issue #1256, opened 2023-05-18
[bug] RainbowKit CSS styling not loading when using 'self' in Content Security Policy on Vercel


ahkill:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### RainbowKit Version

^1.0.0

### wagmi Version

^1.0.1

### Current Behavior

Hello team,

I'm currently experiencing an issue with CSS styling in RainbowKit when deploying on Vercel. I've specified the Content Security Policy (CSP) to 'self' in my vercel.json file. When I run the app on my local machine, the CSS styling loads as expected. However, when I deploy the app on Vercel, RainbowKit's CSS styling does not load.

As a workaround, I found that adding 'unsafe-inline' to the CSP in the vercel.json file allows the CSS to load correctly. However, this option is not ideal due to potential security risks associated with inline styles.

### Expected Behavior

Regardless of the CSP restrictions in vercel.json, all the styles should show up

### Steps To Reproduce

For your convenience, I have created a [minimal reproducible repository](https://github.com/ahkill/rainbow) and a [preview link](https://rainbow-tan.vercel.app/) showcasing the issue.

Steps to reproduce: I have already reproduced, you can clone it for further experimenting
How can i notice the bug: Goto the vercel deployed URL of this repo and you will notice that styles of the connect button and the wallet modal are off (console says CSS variables from rainbowkit package are undefined)

Github repo: https://github.com/ahkill/rainbow
Vercel Link: https://rainbow-tan.vercel.app/

### Link to Minimal Reproducible Example (CodeSandbox, StackBlitz, etc.)

https://github.com/ahkill/rainbow

### Anything else?

<img width="1194" alt="Screenshot 2023-05-18 at 13 21 51" src="https://github.com/rainbow-me/rainbowkit/assets/129762645/137744c9-d830-469a-b63c-7cb532cad6a6">

DanielSinclair:
Hey @ahkill. Thanks for reporting. This is quite tough; it seems like some but not all of the RainbowKit component's styling is hitting this unhappy path. I found this [Understanding and implementing a proper Content Security Policy with Next.js and Styled Components](https://reesmorris.co.uk/blog/implementing-proper-csp-nextjs-styled-components) article to be quite helpful in describing the issue and possible solutions. Will need to keep investigating this, and see if this issues has been described in our upstream styling dependencies like `vanilla-extract`.

WebDev-Akhil:
@DanielSinclair 

[...]

PR #2325, opened 2025-02-19
example: with react router

jgresham:
With Remix and React Router semi-merging, I thought it would be nice to have an updated example using the [recommended](https://remix.run/blog/react-router-v7) React Router moving forward.

Potential issues with this example:
1. `import 'node_modules/@rainbow-me/rainbowkit/dist/index.css';` is required in the component that uses `<ConnectButton/>` and perhaps there is a better fix for the styles that are not being applied without it. (related issue https://github.com/rainbow-me/rainbowkit/issues/1256)

Please let me know if a changset is required or if I messed up the pnpm workspace which I am new to.


DanielSinclair (approved):
Thank you!

jgresham:
> Thank you!

thank you!!

PR #2331, opened 2025-02-20
fix: avoid creating a new WC connector on page reload (when already connected)

### LINKED ISSUE #2232: [bug] WalletConnect connector creates multiple instances and gets unsynced from the app

### RainbowKit Version

2.2.0

### wagmi Version

2.12.25

### Current Behavior

After connecting to the WalletConnect connector and reloading the app, a second instance of the connected connector is created, which prevents the wallet from notifying the app about changes, and eventually throwing a `Error: No matching key. history: <number>` error.

I double checked and this is not an issue if I directly connect through Wagmi's WalletConnect connector. I also tried connecting through WC with a Safe Wallet and a Metamask wallet and could reproduce the problem in both cases.

### Expected Behavior

The WalletConnect connection should keep working between reloads of the app, or worst case scenario, it should automatically disconnect on reloads to prevent the bad UX for the user using the app

### Steps To Reproduce

- Open the Minimal Reproducible Example provided
- Update the `projectId` for a valid WC project ID
- Open the preview in a new tab (top-right corner). It's important to open the app in a full tab to reproduce the issue
- Click the connect button and connect through WalletConnect. In the console you can see that it's connected to a single WC instance
- Reload the tab
- Open the console and you will now see that the app is connected to 2 instances of WalletConnect
- Switch the network through the `ConnectButton`, confirm in your wallet and you will notice that the app will not be notified about the successful action, then you will be able to see the `No matching key. history` error in the console

This error seems to affect all messages sent from the wallet to the app, I've also experienced it with transaction statuses on a larger app, where the wallet never notifies the app about confirmed, successful or transactions with errors

https://stackblitz.com/edit/new-wagmi-bfc6lr?file=src%2Fwagmi.ts%3AL7

### Anything else?

_No response_

agualis:
Fixes https://github.com/rainbow-me/rainbowkit/issues/2232

Uses localStorage to know if there is an existing Wallet Connect connection after reloading the page. 
If yes, it avoids creating a new Wallet Connect connector (`{ ...config, showQrModal: false }`) which was leading to broken connection (with `Error: No matching key. history: <number> error.`).

The bug could be reproduced in the linked issue or in the the default rainbowkit example (Running `pnpm run dev` and following the reproduction steps in the linked issue).

### DIFF
--- a/.changeset/warm-cooks-cover.md
+++ b/.changeset/warm-cooks-cover.md
@@ -1,0 +1,5 @@
+---
+"@rainbow-me/rainbowkit": patch
+---
+
+fix: avoid creating a new WC connector on page reload (when already connected to WC)
--- a/packages/rainbowkit/src/components/RainbowKitProvider/RainbowKitProvider.tsx
+++ b/packages/rainbowkit/src/components/RainbowKitProvider/RainbowKitProvider.tsx
@@ -31,7 +31,7 @@ import { WalletButtonProvider } from './WalletButtonContext';
 import { useFingerprint } from './useFingerprint';
 import { usePreloadImages } from './usePreloadImages';
 import { clearWalletConnectDeepLink } from './walletConnectDeepLink';
-import { clearWalletConnected } from './walletConnectConnectionStatus';
+import { clearWalletConnectConnected } from './walletConnectConnectionStatus';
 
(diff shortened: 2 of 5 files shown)


0xGeegZ (approved):
LGTM

PR #2359, opened 2025-04-04
feat: replace qrcode with cuer package

DanielSinclair:


talentlessguy:
based PR, `qrcode` pulls a bunch of deps that are intended for server-side use

PR #2361, opened 2025-04-06
Add konkretWallet

Kontanten:
Add connector for the [Konkret Wallet](https://codeberg.org/konkret/konkret-wallet) ([GH](https://github.com/konkret-wallet/konkret-wallet)) MM fork.


Kontanten:
> @Kontanten It is a little premature to support this wallet with a connector, but as long as it supports EIP-6963, a branded button will appear in the connect interface when the wallet is installed.

That's reasonable!

Hopefully there will be reason to revive this at a later point, then.

PR #2373, opened 2025-04-12
Update provideRainbowKitChains.ts

jill6666:
Rename the BSC chain name as it has been updated in the definition of viem.

https://github.com/wevm/viem/blob/main/src/chains/definitions/bsc.ts


DanielSinclair:
@jill6666 We still currently prefer BSC, short for BNB Smart Chain

PR #2625, opened 2026-01-24
fix: fix browser freeze with multiple wallet extensions installed

### LINKED ISSUE #2534: [bug] rabbyWallet connector causing crash

- [x] I have searched the existing issues

### RainbowKit Version

2.2.8

### wagmi Version

2.17.5

### TLDR

There is a bug related to `rabbyWallet` connector ([here](https://github.com/rainbow-me/rainbowkit/blob/775e83f4ccb37e962e313b74d45bf192f2839548/packages/rainbowkit/src/wallets/walletConnectors/rabbyWallet/rabbyWallet.ts#L7)) that causes the page to crash/be stuck on an infinite loop

### Resources
- github: I have a public repo setup here where you can reproduce the issue locally: https://github.com/xtbase/test-rb

- sample vercel deployments

- broken (uses rabby connector, notice that increment button doesn't work): https://test-rb-kappa.vercel.app/
- working (doesn't use rabby connector, increment works fine): https://test-rb-kappa.vercel.app/no-rabby

### Expected Behavior

There should not be any crashes when using `rabbyWallet` connector, suspect it's due to some injectedProvider issues with other injected wallets such as metamask

### Steps To Reproduce

1. Have these extensions installed - coinbase wallet + phantom + metamask
2. When setting up `connectors` to pass into wagmi, use `rabbyWallet` connector

```
const connectors = connectorsForWallets(
  [
    {
      groupName: 'Recommended',
      wallets: [
        ...,
        rabbyWallet,
      ],
    },
  ],
);

const wagmiConfig = createConfig({
  connectors,
  ...,
});

```
3. load the page, see that everything is unresponsive and you have to manually close the tabs

https://test-rb-kappa.vercel.app/

### Anything else?

_No response_

Yeonny0723:
Fixes #2534

## Summary

There is an issue that the browser freezes when multiple wallet extensions (e.g., MetaMask, Phantom, Coinbase Wallet) are installed simultaneously. 

I looked into the browser call stack during the freeze and identified an infinite recursion loop between `getAccounts` and the `onConnect` event.

evnetloop

It looks like having multiple extensions installed causes them to fight for the provider injection.
I noticed in the call stack that registerProviders (from the injected script) gets triggered repeatedly. Since this emits a connection event every time, RainbowKit responds by calling getAccounts again, creating an infinite loop that freezes the app.

## The Fix

I modified the `getAccounts` implementation in `getInjectedConnector` to prevent concurrent calls.
- AS-IS> Every `onConnect` event triggered a new async call to `baseConnector.getAccounts()`.
- TO-BE> If a request is already in flight, the connector now returns the existing promise instead of starting a new one in order to prevent the event loop while ensuring the app still receives the account data.

## Demo
* Before change (freeze on initial load):
before

* After change:
after

---

## PR-Codex overview
This PR enhances the `getInjectedConnector` functionality by implementing a deduplication mechanism for `getAccounts` calls, ensuring that multiple requests do not lead to redundant operations. It also includes tests to verify this behavior and error handling.

### Detailed summary
- Refactored `getInjectedConnector` to use a base connector for `getAccounts`.
- Introduced a promise (`getAccountsPromise`) to prevent duplicate `getAccounts` calls.
- Added tests for:
  - Deduplication of concurrent `getAccounts` calls.
  - Error handling for `getAccounts` and promise clearing.
- Updated test cases to mock Ethereum provider behavior.

> ✨ Ask PR-Codex anything about this PR by commenting with `/codex {your question}`

### DIFF
--- a/packages/rainbowkit/src/wallets/getInjectedConnector.test.ts
+++ b/packages/rainbowkit/src/wallets/getInjectedConnector.test.ts
@@ -1,10 +1,26 @@
-import { describe, expect, it } from 'vitest';
+import { describe, expect, it, vi, beforeEach } from 'vitest';
+import { mainnet } from 'wagmi/chains';
 import {
   getInjectedConnector,
   hasInjectedProvider,
 } from './getInjectedConnector';
 
+const mockConfig = {
+  chains: [mainnet],
+  emitter: {
+    on: vi.fn(),
+    off: vi.fn(),
+    emit: vi.fn(),
+    listenerCount: vi.fn(() => 0),
+  },
+  storage: undefined,
+} as any;
+
 describe('getInjectedConnector', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
   it('only rainbow provider', () => {
     window.ethereum = { isMetaMask: true, isRainbow: true };
     const connector = getInjectedConnector({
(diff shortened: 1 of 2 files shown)


DanielSinclair:
Thank you for diagnosing and proposing a fix for this! I've built on your work in #2661 with a slightly different solution that solves the broader root issue.

PR #2687, opened 2026-08-03
fix: cap recent completed transactions at 10

grandpig:
## Summary

- Fix an off-by-one error in the transaction store
- Ensure that at most 10 completed transactions are retained
- Keep pending transactions outside the completed transaction limit
- Add a regression test

## Root cause

The existing filter uses a post-increment with `<=`, which allows the 11th completed transaction to pass:

```ts
completedTransactionCount++ <= MAX_COMPLETED_TRANSACTIONS
```


grandpig:
@DanielSinclair Hi! I noticed you recently worked on the transaction store in #2670. This is a small off-by-one fix in the same area that caps completed recent transactions at 10, with a regression test included. Would you mind taking a look when you have a chance? Thanks!

grandpig:
@DanielSinclair Could you please take a look when you have a chance? Thanks!

PR #2701, opened 2026-09-03
fix: give account modal an accessible name

### LINKED ISSUE #2700: Account modal has no accessible name (aria-labelledby points to a nonexistent id)
## Description

The account modal's dialog (opened by clicking the connected account button) sets
`aria-labelledby="rk_account_modal_title"`, but no element inside the modal has
that id. The dialog therefore has no accessible name: a screen reader announces
it only as "dialog", with no indication of what it is.

## Version

`@rainbow-me/rainbowkit` 2.2.11

## Steps to reproduce

1. Run RainbowKit's own example app (or any app using `ConnectButton`) and connect a wallet.
2. Click the connected account button to open the account modal (the one showing address, balance, "Copy Address" and "Disconnect").
3. Inspect the dialog element in devtools, or listen with a screen reader.

## Expected behavior

The dialog announces a name, e.g. "Account" or the connected address, the way the connect modal announces "Connect a Wallet".

## Actual behavior

The dialog has `aria-labelledby="rk_account_modal_title"`. No element in the DOM has `id="rk_account_modal_title"`. A screen reader announces the dialog with no name. This can be confirmed without a screen reader: `document.getElementById("rk_account_modal_title")` returns `null` while the modal is open.

For contrast, the connect modal works correctly: it sets `titleId = "rk_connect_title"` and renders an element with `id="rk_connect_title"`, so its `aria-labelledby` resolves.

## WCAG / rule

- WCAG 4.1.2 Name, Role, Value
- axe: `aria-valid-attr-value`, and `aria-dialog-name` under an axe run configured to include that check
- IBM Equal Access: `aria_id_unique`

## Source

`Dialog` (`src/components/Dialog/Dialog.tsx`, compiled at `dist/index.js` around the `function Dialog({ children, onClose, open, titleId })` definition) takes a `titleId` prop and sets `"aria-labelledby": titleId` on the dialog element.

`AccountModal` (`src/components/AccountModal/AccountModal.tsx`, compiled at `dist/index.js`, function `AccountModal({ onClose, open })`) does:

```
const titleId = "rk_account_modal_title";
...
React35.createElement(Dialog, { onClose, open, titleId }, ... React35.createElement(ProfileDetails, {...}))
```

It passes `titleId = "rk_account_modal_title"` to `Dialog`, but its child content is `ProfileDetails`, which never renders anything with that id.

`ProfileDetails` (`src/components/ProfileDetails/ProfileDetails.tsx`, compiled at `dist/index.js`, function `ProfileDetails({ address, ensAvatar, ensName, balance, onClose, onDisconnect })`) declares its own, unrelated title id:

```
const titleId = "rk_profile_title";
```

and renders it on two `<h1>` elements (the account name and, when a balance is present, the balance), each with `id: titleId`, i.e. `id="rk_profile_title"`. Neither element carries `id="rk_account_modal_title"`, so `AccountModal`'s `aria-labelledby` never resolves.

By contrast, `ChainModal` and the connect flow follow the pattern correctly: each declares a `titleId` (`"rk_chain_modal_title"`, `"rk_connect_title"`) and renders an element with that exact id inside its own content, so their dialogs do get a name. `AccountModal` is the outlier because it delegates its body to `ProfileDetails`, a component that hardcodes its own id instead of accepting the parent's `titleId`.

## Suggested fix

Have `ProfileDetails` accept the id to use for its title (or render `id="rk_account_modal_title"` when used from `AccountModal`) instead of hardcoding `rk_profile_title`, so the id `AccountModal` declares for `aria-labelledby` is the one that actually gets rendered. Alternatively, point `AccountModal`'s `Dialog` at `rk_profile_title` directly.

---

Found while auditing wallet-connect kits for accessibility. A reproduction script is available on request.

devtechedge:
## Summary

Fixes #2700.

The account modal dialog sets `aria-labelledby="rk_account_modal_title"`, but `ProfileDetails` hardcoded `id="rk_profile_title"` on its headings. The labelled-by id never resolved, so the dialog had no accessible name.

## Changes

- Pass `titleId` from `AccountModal` into `ProfileDetails` so the heading id matches the dialog, same pattern as `ChainModal` / the connect modal.
- Keep the id on the account-name heading only. The balance line was sharing the same id, which is invalid HTML and breaks `aria-labelledby` when a balance is shown.

## Test plan

- [ ] Open the example app, connect a wallet, click the connected account button.
- [ ] `document.getElementById('rk_account_modal_title')` returns the account-name heading while the modal is open.
- [ ] Dialog announces a name (account / ENS), not a nameless "dialog".
- [ ] Balance line still renders; it no longer duplicates the heading id.

### DIFF
--- a/.changeset/account-modal-accessible-name.md
+++ b/.changeset/account-modal-accessible-name.md
@@ -1,0 +1,5 @@
+---
+"@rainbow-me/rainbowkit": patch
+---
+
+Fixed the account modal dialog having no accessible name. `aria-labelledby` now points at the profile heading that actually renders, and the heading id is no longer duplicated on the balance line.
--- a/packages/rainbowkit/src/components/AccountModal/AccountModal.tsx
+++ b/packages/rainbowkit/src/components/AccountModal/AccountModal.tsx
@@ -36,6 +36,7 @@ export function AccountModal({ onClose, open }: AccountModalProps) {
               balance={balance}
               onClose={onClose}
               onDisconnect={disconnect}
+              titleId={titleId}
             />
           </DialogContent>
         </Dialog>
--- a/packages/rainbowkit/src/components/ProfileDetails/ProfileDetails.tsx
+++ b/packages/rainbowkit/src/components/ProfileDetails/ProfileDetails.tsx
@@ -26,6 +26,7 @@ interface ProfileDetailsProps {
   balance: ReturnType<typeof useProfile>['balance'];
   onClose: () => void;
   onDisconnect: () => void;
+  titleId: string;
 }
 
 export function ProfileDetails({
@@ -35,6 +36,7 @@ export function ProfileDetails({
   balance,
   onClose,
   onDisconnect,
+  titleId,
 }: ProfileDetailsProps) {
   const showRecentTransactions = useContext(ShowRecentTransactionsContext);
   const { i18n } = useContext(I18nContext);
@@ -65,7 +67,6 @@ export function ProfileDetails({
   const displayBalance = ethBalance
     ? abbreviateETHBalance(Number.parseFloat(ethBalance))
     : undefined;
-  const titleId = 'rk_profile_title';
   const mobile = isMobile();
 
   return (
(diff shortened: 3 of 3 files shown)


devtechedge:
Hi, just checking in on this one.

It gives the account modal an accessible name for assistive tech.

Open since 3 September with green checks and no conflicts. Happy to adjust anything.

